PR #16, opened 2015-12-01
Adaptive precision tracker

ofloveandhate:
I am requesting a pull from my AMPTracker release branch.  

the code represented in this request is a significant step forward in Bertini2.  It allow us to use the previously accepted correct / predict codes in a tracker loop, which automatically adjusts precision as needed. 

the AMPTracker class inherits from a base Tracker class, and uses polymorphism to allow us to easily write new trackers, such as double precision and fixed multiple precision.  

the tracker has furthermore been used by Tim in his own fork of Bertini2, in conjunction with a new Endgame class, and an implementation of the power series endgame.  

This PR is huge.  There's no way around that.  However, after this request, our PR sizes should shrink dramatically, as the features being added will not be so central.

I have paid attention to the documentation for this class, and believe that the style and content is appropriate.  Tim was able to integrate the Tracker into his Endgame without consulting me -- just the documentation.  

I hereby request formal review, and hope for swift acceptance.


ofloveandhate:
after viewing the commit history, i discover that the same issue Jeb has in his current pull request is in mine -- the commits from my previously resolved pull request are replayed in this PR.  i will work on fixing this immediately.

ofloveandhate:
this is attempting to pull into bertiniteam/b2/master, rather than develop, which is wrong.   i am closing this pull request.

PR #21, opened 2016-02-15
added m4 autoconf macro for finding eigen.

ofloveandhate:
removed eigen3/ from #include paths throughout bertini2.

we were previously looking for eigen in the wrong place, and need to strip a level off the include paths in Bertini2's core.  the submitted m4 macro, called from configure.ac, does just that.  the user can specify the location of their Eigen installation, or let configure attempt to find it itself.


ofloveandhate:
i meant to PR into develop, not master.  my bad.

PR #22, opened 2016-02-15
added m4 autoconf macro for finding eigen.

ofloveandhate:
removed eigen3/ from #include paths throughout bertini2.

we were previously looking for eigen in the wrong place, and need to strip a level off the include paths in Bertini2's core. the submitted m4 macro, called from configure.ac, does just that. the user can specify the location of their Eigen installation, or let configure attempt to find it itself.

### DIFF
--- a/core/configure.ac
+++ b/core/configure.ac
@@ -85,10 +85,7 @@ AC_SEARCH_LIBS([cos], [m], [], [
   ])
 
 # look for a header file in Eigen, and croak if fail to find.
-AC_LANG_PUSH([C++])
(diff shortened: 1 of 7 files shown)


ofloveandhate:
i vote for the immediate acceptance of this pull request.  this has been incorporated into my develop branch anyway, and i have successfully used it on every computer i've tried it on.  the only thing it's missing is version checking, and that can be added later.

dbates1:
I agree.  No need for a review of something like this.

bertiniteam:
accepted

PR #34, opened 2016-06-21
Documentation & Test Case Revisions for Endgames

timothodge:
All documentation has been updated for Cauchy and PowerSeries Endgames.
     - Test cases that were flawed have now been updated to test for consistency.
     - Temporary variable rand_vector has been replaced with a permanent variable that is set once.


bertiniteam:
accepted.  minor revisions to come as i get this into master.

PR #35, opened 2016-07-01
removed copy-and-swap, increased precision uniformity throughout

ofloveandhate:
this set of commits brings several fixes to the table.  

* copy-and-swap for bertini::complex was super slow.  every copy means TWO malloc's through mpfr, and TWO free's.  measured performance improvement was 23% of runtime for me on my mac, with expression templates for real floats turned on, in the endgame test suite.
* fixes for some problems manifesting in boost 1.60.0_2, related to expression templates.


bertiniteam:
checked with 

* redhat linux, boost 1.58 unpatched, boost 1.61, and eigen 3.2.8.  
* OSX, patched 1.58, 1.60.0_2, and eigen 3.2.8

am accepting.

PR #39, opened 2016-08-03
Hotfix for the MPFR parser

jbcolli2:
The current MPFR parser has problems with reading numbers with scientific notation.  In particular, reading integers that use scientific notation, and reading numbers with all digits after the decimal with scientific notation.

 Also, I have moved the MPFR parser to a separate struct MPParserRules in the file function_parsing.hpp.  This allows that parser to be accessed by other parsers that may need to read MPFR numbers.

### DIFF
--- a/core/include/bertini2/function_tree/function_parsing.hpp
+++ b/core/include/bertini2/function_tree/function_parsing.hpp
@@ -370,71 +370,6 @@ namespace bertini {
 
 
 
-//			integer_string_.name("integer_string_");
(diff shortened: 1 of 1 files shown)


ofloveandhate:
there are some comments of old code left in the parsers.  will you please delete these?

jbcolli2:
Done.

ofloveandhate:
hey jeb, will you re-open PR into master, rather than dev, for hotfix?  i want to get better with our flow, and i think hotfix --> master is better than develop.

PR #42, opened 2016-08-11
Parsing config settings

jbcolli2:
A set of parsers for each configuration struct that parse the corresponding settings in an input file.  Also, a larger function that takes as template parameters each of these structs, as well as the config portion of an input file.  The function then returns a tuple containing all the structs.

All new code in the file config_parsing.hpp, with new tests in settings_test.cpp.


ofloveandhate:
i believe that bertini1 allows space between the name of the setting and the colon.  can you modify your parsers to allow this?

jbcolli2:
Thanks for letting me know, changes have been made.

ofloveandhate:
thanks!

PR #43, opened 2016-08-11
Endgames generalized to handle nonzero target time.

timothodge:
* Cauchy endgame can now be used to approximate solutions for t not equal to 0
* Power series endgame can now be used to approximate solutions for t not equal to 0
     * Functions that implicitly use target_time = 0 now have test cases to check if they are able 
        to handle when the target time is nonzero.

### DIFF
--- a/core/include/bertini2/tracking/base_endgame.hpp
+++ b/core/include/bertini2/tracking/base_endgame.hpp
@@ -224,6 +224,7 @@ namespace bertini{
 					We track then to the next_time and construct the next_sample.
 
 				\param start_time The time value at which we start the endgame. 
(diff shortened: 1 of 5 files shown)


ofloveandhate:
at cauchy line 361,

`if(target_time != BCT(0,0) && next_time != starting_time)`

surrounds an addition.  i'm not sure this conditional simplifies anything.  what's your opinion?

timothodge:
Agreed that it does not simplify anything. I have removed the conditional and moved the addition into the above next time calculation. See the newest commit for how this was done.

ofloveandhate:
thanks!

PR #53, opened 2016-09-08
Auto Retrack of crossed paths

jbcolli2:
Check is done for crossed paths by determining if two paths are the same at the endgame boundary.  Those paths are then rerun with a reduced tracking tolerance.


jbcolli2:
Dan,
   I believe the proper changes have been made, feel free to review it again whenever you have time.

PR #65, opened 2017-01-18
zerodim, nid, re-organization

ofloveandhate:
this pull request brings some big nice changes to bertini2:

* `std::make_shared<>` has been largely replaced with calls to `MakeVariable`, etc, allowing for future implementation of a management scheme for things like nodes, etc, using the pools.
* zerodim algorithm is looking pretty good.  has methods for accessing the settings, etc. 
* basic skeleton for NID is implemented.
* witness set implemented, with policies for storing the points, etc.


ofloveandhate:
i am accepting my own PR ;)

PR #66, opened 2017-01-26
start system restructure, add of m-hom skeleton

ofloveandhate:
this PR brings a better structure of the system folder, and a place for the m-hom start system, for tim and karleigh


ofloveandhate:
self-approving

PR #84, opened 2017-09-13
compile fixes, boost version increase, and events for endgames

ofloveandhate:


ofloveandhate:
this built cleanly on four machines.  two osx, and two ubuntu's.

ofloveandhate:
note that ubuntu 16.04 can no longer just get boost from apt, as the apt-distributed version for 16.04 is 1.58 ... and the min version of boost for bertini2 is now 1.62.  there is good reason: 1.62 brings `isnan`, and 1.61 brought expression template fixes that eigen needed to be able to work in bertini2.

ofloveandhate:
this PR also brings events to the endgames, yay!  now we can observe what's happening at various points.  

power series still needs its notification points written.  this will come very soon.

PR #88, opened 2017-09-19
M-Hom Start System

jbcolli2:
This pull request includes the m-homogeneous start system.  To create this start system a new node type, LinearProduct, was created which represents a linear product for a particular variable group.  Tests were added into the b2_class_tests for the LinearProduct node type.


ofloveandhate:
thanks so much for the pull request.  i'll get to reviewing it soon! 🎴

ofloveandhate (approved):
I did a test merge on my latest develop branch, and this merges beautifully.  no conflicts, nothing breaks.  

a quick set of commits should follow this resolution, which expose the m-hom start system to PyBertini, and prepare a tutorial using it to solve something.  some quick sampling of the possible variable groupings, programmatically, would be spectacular.  we can now incorporate this easily into bertini, if we wish.

PR #112, opened 2018-01-16
python documentation upgrade

ofloveandhate:
this pull request represents the addition of Sphinx to PyBertini is its documentation system. 

Inspired by GalSim, particularly [issue 160](https://github.com/GalSim-developers/GalSim/issues/160), I completely rewrote the way the bindings are structured.  its now far more canonical.  
* `_pybertini` is the internal C++ side, raw generated boost bindings
* `pybertini` is the pure python side, wrapping the generated library into a nice package, with documentation done in sphinx
* you can install the pure python bindings with `pip install .`, which is cool.  before, there wasn't an easy way to install them

additionally, 
* there are now two tutorials, 
  * one of which loops to track paths to nonsingular endpoints
  * the other constructs and evaluates the cyclic-n system, for n = 10.  it's super easy to formulate systems using PyBertini, with programmatic variable creation and convenient ways to loop and define systems with structure.
* i made many improvements to the exposure of multiprecision numbers, and in other places throughout the library
* all tests work again.  pybertini needs more testing, but that's always the case, ha.


ofloveandhate:
i just checked out a fresh copy and everything went smooth, both c++ and python

PR #121, opened 2018-02-01
fix broken `pip install .` for pybertini

ofloveandhate:
added missing package_dir


ofloveandhate:
self-accepting. 

may it break nothing and benefit many  💮

PR #140, opened 2018-02-11
logging and pybertini fixes

ofloveandhate:
now can adjust the logging destination and level at runtime from both c++ and python.  huzzah.  🎊


ofloveandhate:
added a tutorial on the endgames, specifically using griewank osborne and the cauchy endgame

PR #141, opened 2018-02-12
remove temp files

jbcolli2:
Update to performance numbers, with temporary files removed and .gitignore updated.


ofloveandhate:
thanks!!!

PR #183, opened 2023-07-12
Feature/python expose zero dim

ofloveandhate:
this PR represents exposing zero dim solvers to Python, and a bunch of other related changes, including switching from minieigen to EigenPy.


ofloveandhate:
i do note that this leaves a bunch of things incomplete.  but i want to get the switch to eigenpy into develop asap.

PR #185, opened 2023-07-13
Feature/fix eigen pass by reference

ofloveandhate:
an attempt to get functions which return by pass by reference to be callable using numpy objects from python.

the workaround is super hacky.  i hate it.  the real solution is to actually use `Eigen::Ref` throughout the core.  please make that change (it's not trivial).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,7 +37,7 @@ python/build/
 # Python files
 *.pyc
 
-
+python/pybertini_*.log
 
 *.dirstamp
 *.deps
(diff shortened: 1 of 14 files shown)


ofloveandhate:
the unit tests in python all pass now.  there's probably a lot of uncovered code, though, so my guess is there's still a fair amount of functions that need to be wrapped.  (they wouldn't need wrapping if they use used `Eigen::Ref` to start with.

ofloveandhate:
i do want to add a note about Eigen::Ref.  it doesn't allow resizing.  So that's kinda fucked.  But pass-by-reference DOES.

Issue #192, opened 2023-11-29
Compile fails on Ubuntu 20.04 LTS


BrentBaccala:
System is Ubuntu 20.04 LTS.  Bertini 2 is cloned from github.

`configure` runs fine.

Compilation in `b2/core` fails.

It seems to require `bmp::variable_precision_options`, which appears to have been introduced in Boost 1.77.

The configure script only checks for Boost 1.71, and the compilation instructions suggest that Ubuntu 20.04 LTS (with Boost 1.71 from apt) is a usable build system.

Probably the build instructions are outdated.

BrentBaccala:
Even the recent Ubuntu 23.10 release (Mantic Minotaur) only provides Boost 1.74.

So I guess I have to compile Boost myself if I'm on an Ubuntu system, right?

ofloveandhate:
Unfortunately, yes.  I wish the boost versions provided by apt were more current.  And, I found a bug in 1.82 that's fixed in 1.83, so 1.77 isn't even enough.  (It's a minor bug, but it's there).

I am sorry the build instructions are out of date.  I find it nearly impossible to keep up, as I teach ≥12 credits each semester, plus service.  Plus, the general shit environment for trans people in the US has me deeply depressed.  Working on b2 is incredibly satisfying, but I rarely find time or mental state for it.

I *do* have a student working on re-writing the build system from the autotools to CMake, which should dramatically improve the building process, and enable us to distribute using conda.  So that's nice.  

In general, though, bertini 2 is NOT stable, ready-to-rock software.  It has some sweet features and I'm quite proud of it.  But alas, I ended up at an undergraduate liberal arts school with a heavy teaching load and little time to work on it.

I hope this was helpful.

ofloveandhate:
I regard b2 as an experimental piece of software.  I hope you enjoy what experience you have with it.  Sorry it's not what it's supposed to be.  It's missing the numerical irreducible decomposition, and is not parallelized.  Bertini 1 is much more complete.  If you need a reliable system right now, please consider using Bertini 1.

ofloveandhate:
One more: 

I plan to add the NID to b2 in AY24-25.  I have a year of sabbatical planned and included NID in the proposal, and I think I will be very successful.

ofloveandhate:
ok, one more.  

[...]

PR #193, opened 2023-12-07
Increase required Boost version from 1.71 to 1.83 in configure.ac

BrentBaccala:
see bertini2 issue #192


ofloveandhate:
Thank you.

PR #197, opened 2024-04-24
Can now install with CMake and pip

ThisIsNotANamepng:
As a result of my research with Dr. amethyst, bertini2 can now be built with CMake, and pybertini can then be installed with pip.


ofloveandhate:
There's a bit to do yet, but this worked on my M1 Mac.  The gaps I see right now:
* some of the test suites in the core are not built.  the sources are listed, but the executables are not created.  should be simple to fix.
* 🤔

ofloveandhate:
* all core tests are represented
* built easily on my m3 mac just now as-is.

PR #220, opened 2026-05-18
👷🏻 strip eigenpy symbols before packing wheels

ofloveandhate:
this gets the wheels to less than 100 megs.


ofloveandhate:
linux wheels are still rather large due to eigenpy.  but that's ok for now.

PR #225, opened 2026-05-19
v.2.0.2 to main

ofloveandhate:
solves CI problems with documentation generation as part of deployment


ofloveandhate:
I'm accepting this without a dev release, as it's really only the build matrix and eigenpy version that changed, together with workflow improvements.

Issue #70, opened 2017-03-30
cauchy eg doesn't respect non-zero target time


ofloveandhate:
in the function `RatioEGOperatingZoneTest()`, the line
```
if(norm(times.front()) < GetCauchySettings().ratio_cutoff_time)
```
 is faulty, it should be the difference between the front time and the target time.

ofloveandhate:
fixed by 057b4baaa0e0bf8f3c73d8fa06514417821ad4ea 
thanks tim!

Issue #49, opened 2016-09-01
__getitem__ demands a Python long integer


jacobbond:
When trying to get an item out of a VectorXmp, `__getitem__` demands a Python long integer.  It doesn't actually affect functionality and it can be worked around in code with `long(str(index))`.

```
sage: y_end[1]
---------------------------------------------------------------------------
ArgumentError                             Traceback (most recent call last)
<ipython-input-53-4e4a51777556> in <module>()
----> 1 y_end[Integer(1)]

ArgumentError: Python argument types in
    VectorXmp.__getitem__(VectorXmp, Integer)
did not match C++ signature:
    __getitem__(Eigen::Matrix<bertini::complex, -1, 1, 0, -1, 1>, long)
sage: y_end[1L]
(0.0000000000000000e+00, 0.0000000000000000e+00)
```

ofloveandhate:
is the (a) fix to provide functions accepting int's?  

can you confirm this issue also exists with VectorXd?

jacobbond:
Yes, it does happen with VectorXd:

```
sage: floats = [float(n) for n in range(5)]
sage: VectorXd(floats)[0]
---------------------------------------------------------------------------
ArgumentError                             Traceback (most recent call last)
<ipython-input-32-d82c085cbe41> in <module>()
----> 1 VectorXd(floats)[Integer(0)]

ArgumentError: Python argument types in
    VectorXd.__getitem__(VectorXd, sage.rings.integer.Integer)
did not match C++ signature:
    __getitem__(Eigen::Matrix<std::complex<double>, -1, 1, 0, -1, 1>, long)
sage: VectorXd(floats)[0L]
0j
```

I don't anticipate ever needing a vector longer than `INT_MAX`, but I can't speak for others.  I did find that Python's `int`s are implemented using `long`s ([here](https://docs.python.org/2/library/stdtypes.html#numeric-types-int-float-long-complex)), so there is kind of a disconnect between what Python might pass as an `int` and what C++ can accept as an `int`.

ofloveandhate:
this is a [minieigen](https://github.com/eudoxos/minieigen) problem.  can you file an issue there, and refer to it from here?

jacobbond:
Filed with [minieigen](https://github.com/eudoxos/minieigen) as [#14](https://github.com/eudoxos/minieigen/issues/14).

ofloveandhate:
[...]

Issue #33, opened 2016-06-03
Multithreading almost certainly broken in OSX


ofloveandhate:
I used the `thread_local` keyword for a static variable in the complex class, to help eliminate temporaries during complex arithmetic.  However, Apple's official Clang compiler does NOT implement this keyword.  Hence, Mac compilation explicitly requires the user to disable thread_local by use of a configure flag (--disable-thread_local).  This leaves the core with a thread-unsafe static variable on Mac.  

As of the time of this writing, threads are not used in the core itself.  However, I believe that multithreading is of critical importance, so fixing this issue is critical in the near future.  Not only this, but since the core is a library, users of the core may write multithreaded client code, and yield race conditions.

It would be awesome for someone to invest the time to more correctly thread-safety-ize statics used to eliminate expensive temporaries in complex arithmetic.  See b2/core/include/bertini2/mpfr_complex.hpp.

ofloveandhate:
i read recently that the new xcode should bring `thread_local` to osx clang finally.  we'll see.

ofloveandhate:
Xcode 8 brings `thread_local` to the table, finally.  so, i am closing this.  people with earlier versions of the compilers provided by Apple can continue to use `--disable-thread_local`, but now those with ≥8 can just `./configure` and be happy.

Issue #31, opened 2016-05-31
Python bindings issue warning for to-python converter


ofloveandhate:
```
In [1]: import pybertini
/usr/bin/ipython:1: RuntimeWarning: to-Python converter for std::shared_ptr<bertini::node::Node> already registered; second conversion method ignored.

```
this appears regardless of build environment.

ofloveandhate:
i have resolved this.

PR #238, opened 2026-06-16
CLI, so many bugfixes, block structured systems, mhom exposed, user homotopy, two-level parallelism...

### LINKED ISSUE #51: system generator
i'd like to request a generator to produce input sections of bertini classic input files, given a bertini::System.

### LINKED ISSUE #71: constructor for total degree is wrong if system is non-affine-based
if the system provided was not formulated using an affine variable group, the formed system will be wrong, or throw, or some other evil thing

### LINKED ISSUE #74: system parser constructs generic power nodes rather than int power nodes
parsing a system from string produces incorrect power nodes.  

you can reproduce this in Python with the following lines of code:
```
from pybertini import *
sys = parse_system('function f1, f2; variable_group x,y,z; f1 = x*y; f2 = x^2*y - z*x;')
sys.differentiate()
sys
```
the resulting operator for the derivative of the x^2 term should just be x, not x^(2-1).

### LINKED ISSUE #87: random number generation needs centralization
random number generation needs a solid interface, with a way to seed, etc, that works across threads and number types.

### LINKED ISSUE #95: Check the wrapper for Reset for generic nodes
i noticed something was up with it.

### LINKED ISSUE #100: document that bertini variable values are random by default
the justification is that for system simplification, it eliminates zeros and ones.  so 0, 1, and nan are out for default values.  hence, random.

### LINKED ISSUE #107: optimize resetting of nodes in function tree?
should all nodes memoize?  currently, they do.  it's probably too much, and wasteful of time.

### LINKED ISSUE #116: add `final` to final types in core
to allow optimizing out certain virtual calls.  of course, profiling is key.

### LINKED ISSUE #123: add str methods for many types missing it in pybertini

### LINKED ISSUE #128: expose m-hom start system to pybertini

### LINKED ISSUE #129: move start system from being under the system stuff

### LINKED ISSUE #130: add tutorial on endgame usage

### LINKED ISSUE #131: add tutorial on how to solve a polynomial system -- start to finish

### LINKED ISSUE #132: user homotopy is missing the link to its start points

### LINKED ISSUE #133: add tutorial on user defined homotopies

### LINKED ISSUE #148: pybertini complexes are not pickleable
```
RuntimeError: Pickling of "_pybertini.multiprec.Complex" instances is not enabled (http://www.boost.org/libs/python/doc/v2/pickle.html)
```
🥗

### LINKED ISSUE #154: precision consistency: Bug with when homogenizing the variables
There is a bug when running the following code in IPython:

```
import pybertini
import copy

gw = pybertini.System()

x = pybertini.Variable("x")
y = pybertini.Variable("y")

vg = pybertini.VariableGroup()
vg.append(x)
vg.append(y)
gw.add_variable_group(vg)

gw.add_function((x+2)*(y-3)**6)
gw.add_function((x**2 - 1)*(y-5))

gw.homogenize()
gw.auto_patch()

t = pybertini.Variable('t')
td = pybertini.system.start_system.TotalDegree(gw)
gamma = pybertini.function_tree.symbol.Rational.rand()
hom = (1-t)*gw + t*gamma*td
hom.add_path_variable(t)



tr = pybertini.tracking.AMPTracker(hom)

start_time = pybertini.multiprec.Complex("1")
eg_boundary = pybertini.multiprec.Complex("0.0")

midpath_points = [None]*td.num_start_points()
for ii in range(td.num_start_points()):
        midpath_points[ii] = pybertini.multiprec.Vector()
        code = tr.track_path(result=midpath_points[ii], start_time=start_time, end_time=eg_boundary, start_point=td.start_point_mp(ii))

```

When this code is run in `python3`, the following error occurs,
```
Assertion failed: ((bertini::Precision(x(0))==DoublePrecision() || bertini::Precision(x(0)) == Precision()) && "precision of input vector must match current working precision of patch during rescaling"), function RescalePointToFitInPlace, file ./include/bertini2/system/patch.hpp, line 392.
Abort trap: 6
```

If the lines
```
gw.homogenize()
gw.auto_patch()
```
are commented out, then no error occurs.

### LINKED ISSUE #156: pi computed inefficiently
throughout the code base, pi should be computed using something other than the used methods (acos(-1))

### LINKED ISSUE #198: computational inefficiency
I found some pessimizations in explicit_predictors.hpp, in `FullStep()`.  A temp vector is set to 0 when it doesn't need to be.  It can just be set to the first of them, and then the loops can run from 1 to bla instead of from 0 to bla.  this is probably a fairly important improvement to make.

### LINKED ISSUE #199: rename stages and variables related to ODE predicting
right now in `explicit_predictors.hpp`, the member variable`s_` is the number of stages, and it's hard to read this code.  There's a lot of `blaref` naming, too.  It's also making this hard.

### LINKED ISSUE #210: i wish i could make a system in Python from a list of functions, or even from a single function

### LINKED ISSUE #228: stringification produces way too many parentheses
For example,

---

```
import bertini
x, y = bertini.Variable(n) for n in ['x','y']
print(x*y)
```

gives 

```
(x*y)
```

which is correct and safe, but also stupid.

---

```
x**2 + y*x**3 - 3
```

gives 

```
(((x^2)+(y*(x^3)))-3)
```

which is definitely too much.  

---

Eliminate the extra parens.  Perhaps start [here](https://stackoverflow.com/questions/1063316/how-to-get-rid-of-unnecessary-parentheses-in-mathematical-expression).  Note that this fix will be correctly done in the C++ core.

### LINKED ISSUE #232: improve CI build time
building in CI Github Actions takes wayyyy too long.  Boost is re-built every time, plus many other pessimizations.  Improve.  See [here](https://markaicode.com/github-actions-cicd-optimization-slow-to-fast/) for inspiration.

### LINKED ISSUE #235: endgames need a genuinely adaptive precision mode
The trackers run in adaptive, but the endgames themselves not.  This causes a bunch of wasted computation doing things in mpfr when they could be done in double.  A full rewrite is demanded.

### LINKED ISSUE #236: make observers from python with callback functions
I don't know what observers i want, and so i should be able to make them at runtime from Python.

### LINKED ISSUE #239: Ship the (non-MPI) bertini CLI in the Python wheel
## Summary

`pip install bertini` ships the `_pybertini` extension and `libbertini2`, but **not** the blackbox CLI executable. Vanilla users (threads-only, single-machine parallelism) must build from source just to get the `bertini` command, which is onerous. MPI parallelism legitimately requires a from-source build against a matching system MPI — but the threaded CLI has no such constraint and should be installable from the wheel onto the environment's `bin/`.

## Current state

- The CLI is the blackbox executable built from `core/src/blackbox/` (`BERTINI2_EXE_SOURCES` in `core/CMakeLists.txt`).
- The only install rule is `install(TARGETS bertini2 …)` — the executable is never installed into the wheel.
- The threads/MPI split already exists: `find_package(MPI)` + `option(ENABLE_MPI … ${MPI_FOUND})`, with all MPI code conditional. A wheel build is MPI-off by construction.
- `pyproject.toml` uses scikit-build-core with `wheel.install-dir = "bertini"`.

## Proposal (the `ruff`/`cmake`/`ninja` pip pattern)

1. Build the executable in the wheel (MPI off) and `install(TARGETS <exe> RUNTIME DESTINATION _bin)` so it lands at `bertini/_bin/bertini`.
2. Add a `[project.scripts]` console-script shim (`bertini = bertini._cli:main`) that `os.execv`s the bundled binary resolved relative to the package. This puts it on PATH cross-platform (incl. `Scripts\bertini.exe` on Windows) and avoids the "scripts-dir binaries don't get RPATH-repaired" problem.
3. Ensure the bundled executable resolves the **vendored** shared libs that `auditwheel`/`delocate`/`delvewheel` bundle for `_pybertini` (`libbertini2`, Boost, GMP/MPFR/MPC): place it in the package tree next to them and set RPATH (`$ORIGIN/../bertini.libs` on Linux, `@loader_path/../.dylibs` on macOS). Reuses the already-bundled libs, so the size cost is ~the executable alone.

## Tasks

- [ ] Enable + install the blackbox exe in the scikit-build CMake path (MPI off)
- [ ] `bertini._cli` exec shim + `[project.scripts]` entry
- [ ] Per-platform RPATH/repair verification (auditwheel/delocate/delvewheel handling the executable)
- [ ] Smoke test: `pip install` the wheel, run `bertini` on Linux/macOS/Windows

### LINKED ISSUE #243: Auto-randomize-and-filter overdetermined inputs in ZeroDim and the CLI
## Summary

Overdetermined systems (N functions, n variables, N > n) with isolated solutions can now be squared up via `System::Randomize` / `RandomizationBlock` (ADR-0025), solved, and their extraneous solutions filtered by re-evaluating the original. Today the user must orchestrate randomize → solve → filter by hand.

## Proposal

Teach `ZeroDim` (and the blackbox CLI) to accept an overdetermined system directly and do the right thing automatically:
1. randomize/square up via the RandomizationBlock,
2. solve the square system,
3. filter extraneous solutions by re-evaluating the original system,
4. return the genuine isolated solutions.

This makes overdetermined input "just work" end-to-end, both from Python and the CLI.

## Notes

- `Randomize` returns a new system and never mutates the caller; the randomization matrix is retrievable for reproducibility/reporting.
- Filtering = re-evaluate the original functions at each candidate and keep the common zeros.

### LINKED ISSUE #247: Reparseable export of systems: Bertini 1 classic format
## Summary

`System` now has a human-facing, per-block `Describe` (terse placeholders by default, full coefficients with `verbose=True`). That print is intentionally **not** reparseable. A separate, reparseable **export** would let users check Bertini 2 results in another NAG system.

## Proposal

A reparseable exporter for block-composed systems, targeting:
- **Bertini 1 classic input format**

Structured blocks would expand to nodes / explicit equations for export. The natural implementation is a per-block emit contract (`EmitClassic`) parallel to the existing per-block `Describe`.

## Notes

- The per-block `Describe` contract is the stepping stone — same dispatch, different output format.
- Out of scope for the human-printing work; tracked here as the follow-on writer.

### LINKED ISSUE #249: Rewrite predictor/corrector: one random direction per track, corrector as a pure kernel
## Summary

Rewrite the predictor/corrector subsystem so that **the random direction used for the condition-number estimate is owned by the track (one direction per whole path), and the corrector is reduced to a pure kernel** `(J, f, probe) -> (Newton step, condition estimate)` with no RNG of its own.

## Motivation

The Newton corrector uses a random probe vector to estimate `‖J⁻¹‖` (and thus the condition number that drives AMP precision decisions). For a solve to be reproducible and for serial and distributed runs to agree bit-for-bit, that probe must be **stable over an entire path track** (start → endgame → target), and drawn from a controlled RNG point.

Two things are currently wrong/awkward here:

1. **Split ownership.** The corrector (`NewtonCorrector`) owns the probe storage; the *tracker* triggers a refresh once per path. This works (see ADR-0024) but is an interim arrangement — the corrector still draws its own randomness (e.g. `rand_temp_` in `ChangeSystem`).
2. **"Per `TrackPath`" ≠ "per path."** A first attempt to refresh the probe inside `Tracker::TrackPath` was wrong and had to be reverted: the endgame issues *many* `TrackPath` calls per point, so refreshing there redraws mid-track, churns the RNG stream, and regressed serial==distributed bit-identity. The correct granularity is once per *path*, triggered by the algorithm.

## Desired end state

- The random direction is **track/tracker-owned**, set once per path and held constant across the entire track including the endgame.
- The corrector becomes a **pure kernel**: given `(J, f, probe)` it returns `(step, condition estimate)` and contains no randomness.
- This removes the last known source of uncontrolled nondeterminism on singular paths (see the companion issue on residual serial≠distributed precision divergence).

## References

- ADR-0024 (condition-probe per-path refresh — documents the current interim split and the "per TrackPath" pitfall).
- The user has flagged a full predict/correct rewrite as on the horizon; this captures the load-bearing requirement it must satisfy.

### LINKED ISSUE #251: Non-mutating (copy-on-write) Simplify for the function tree
## Summary

Provide a **non-mutating (copy-on-write) `Simplify`** for the function tree: one that returns a new, simplified tree and leaves its input untouched.

## Motivation

Today's `Simplify` mutates the tree in place. Because nodes are shared (`shared_ptr` subtrees), an in-place simplify on a shared subtree lets one `System` observe another's tree change. That hazard is exactly why `DefaultAutoSimplify()` had to be set to `false` (see ADR-0011) — auto-simplification could not be left on safely.

A simplify that produces a new tree (sharing untouched subtrees, copying only what changes) would let auto-simplification be re-enabled and keep shared subtrees immutable.

## Scope

- A non-mutating `Simplify` returning a fresh tree; original unchanged.
- Expose the `EliminateZeros` / `EliminateOnes` building blocks (currently internal).
- This is the read-path counterpart to the simplified-by-construction differentiation already in place: the `SimplifiedMult`/`SimplifiedSum` factories from ADR-0011 flatten, fold exact constants, and never emit `Float`s — a non-mutating `Simplify` should reuse the same rules.

## References

- ADR-0011 (simplified-by-construction differentiation; the `DefaultAutoSimplify()=false` watch note explains the in-place mutation hazard this issue removes).

### LINKED ISSUE #253: Refresh solving_at_scale timing tables on real multi-core hardware
## Summary

Regenerate the MPI scaling timing tables in the `solving_at_scale` tutorial on real, dedicated multi-core hardware.

## Why it can't be done on the current dev environment

The dev VM's `hwloc` cannot read `/sys` CPU topology, so `mpirun` only runs under `--oversubscribe --map-by slot` and produces **negative** speedup (e.g. cyclic-6: serial ~300 s, `-n 5` ~500 s) — the vCPUs are burstable/oversubscribed and throttle asymmetrically. Correctness tests are fine (deterministic), but timings are meaningless there.

## Action

- Run on dedicated cores and update the tables in `python/docs/source/tutorials/solving_at_scale.rst`.
- Tooling exists: `tools/update_scaling_timings.py` (pass `--seed` for reproducibility — unseeded gamma variation otherwise masquerades as a perf difference).
- The observed ~3× ceiling on 8 cores (both cyclic-6 and the eigenvalue system) is a **single-straggler-path** bound — one near-singular / ill-conditioned path sets the floor. That's honest to state in the tutorial, and ties into the predictor/corrector + AMP work as a perf target.

### LINKED ISSUE #255: Refactor the observer subsystem: return-value unsubscribe, weak_ptr lifetimes, typed dispatch
## Summary

The observer subsystem works but has four structural problems. A `NotifyObservers` snapshot fix made the current usage **safe**, but the underlying design is still awkward and unsafe-by-convention. This issue tracks a dedicated refactor.

All four points below were verified against current `develop`.

### 1. Self-removal by mutation during dispatch
`AMPObserver::Observe()` calls `t.RemoveObserver(*this)` on the first precision increase (`core/include/bertini2/trackers/observers.hpp:74`) — a fire-once pattern implemented as mutation of the container being iterated. `NotifyObservers` snapshots each observer list before iterating (`core/include/bertini2/detail/observable.hpp:128–147`) precisely to make that safe. The snapshot is a band-aid.

**Better:** `Observe()` returns an `ObserveResult { KeepObserving, Unsubscribe }`; the dispatcher defers removal to after the loop. No snapshot, no possible iterator invalidation.

### 2. Typed subscription and `dynamic_cast` are inconsistent contracts
`SubscribedEventTypes()` (`observer.hpp:81`) declares the exact `type_index`es an observer wants, and the dispatcher routes only exact-type matches. Yet the typed observers' `Observe()` bodies still use `dynamic_cast` as if any event could arrive — **34** `dynamic_cast`s across the tracker/endgame observers. If the subscription list is correct, those casts always succeed and should be `static_cast` / direct dispatch; if the cast is genuinely needed, the subscription list is incomplete.

**Better:** typed observers get an overloaded `Observe(ConcreteEventT const&)`, removing both the `dynamic_cast` and the `SubscribedEventTypes()` boilerplate.

### 3. Raw-reference storage — no lifetime enforcement
Observers are stored as `std::reference_wrapper<AnyObserver>` (`observable.hpp:153`); `AddObserver` takes `AnyObserver&` and stores `std::ref`. Nothing prevents the observer being destroyed while still registered. Current uses are safe only by convention (`ZeroDimSolve` adds before a path and removes after).

**Better:** register as `shared_ptr<AnyObserver>`, store `weak_ptr`; `NotifyObservers` locks each and skips/prunes expired ones lazily. This also makes `RemoveObserver` optional and removes the need for it to be dispatch-safe.

### 4. Event temporaries bound to references
`NotifyObservers(SomeEvent<Emitter>(*this))` binds a temporary to `AnyEvent const&`, alive only for the call. Any observer storing a reference to the event past `Observe()` would dangle. No observer does today, but it's a trap.

**Better:** pass events by value (cheap — events are small), or document explicitly that event references must not outlive `Observe()`.

## Scope / files

`core/include/bertini2/detail/observable.hpp`, `detail/observer.hpp`, `trackers/observers.hpp`, `endgames/observers.hpp`, and `nag_algorithms/zero_dim_solve.hpp` (the add/remove caller). No public Python API impact — this is an internal cluster.

Suggested sequence:
1. Change `Observe()` to return `ObserveResult`; update all observers; remove the in-`Observe()` `RemoveObserver` calls and the `NotifyObservers` snapshot copies.
2. Replace raw-reference storage with `weak_ptr`; prune expired observers lazily.
3. Audit the `dynamic_cast`s — replace with `static_cast`/typed overloads where the subscription guarantees the type.
4. Document (or enforce by value) the `AnyEvent` lifetime rule.

## Tests

- self-unsubscribe via return value;
- expired `weak_ptr` pruning;
- notify-while-removing does not skip a still-subscribed observer.

### LINKED ISSUE #256: Concatenate using cloned systems.
I attempted to shortcut the system creation process by cloning the original system so I don't always have to reassign the variable groups for each new system. However when attempting to concatenate the two systems, who should have identical variable groups of the same ordering, it throws this error: 

RuntimeError: concatenating systems with differing variable orderings

```
import bertini as b2

x, y = b2.Variable('x'), b2.Variable('y')
var_group = b2.VariableGroup([x,y])

orig_sys = b2.System()
orig_sys.add_variable_group(var_group)

new_sys = b2.system.clone(orig_sys) #don't have to setup the system from scratch

#two generic functions
orig_sys.add(2 * x - 7 * y - 1)
new_sys.add(x - 3 * y)

b2.system.concatenate(new_sys, orig_sys) #<- error here
```

### LINKED ISSUE #258: ZeroDimCauchyAdaptivePrecisionTotalDegree not iterable
When create a user homotopy for a slice move, I received the error:
TypeError: 'ZeroDimCauchyAdaptivePrecisionTotalDegree' object is not iterable.

The error traces to this file:
b2/lib/python3.14/site-packages/bertini/nag_algorithm/__init__.py:220, in user_homotopy(homotopy, start_points, target, precision, endgame)

Example:
```
import bertini as b2

x,y = b2.Variable('x'), b2.Variable('y')
vars = b2.VariableGroup([x, y])
#Lemniscate of Gerono system
f = x**4 - x**2 - y**2

#intersection lines from l to m
l = 2 * x - 7 * y - 1
m = x - 3 * y

#system setup
sys = b2.System()
sys.add_variable_group(vars)

start_slice = b2.system.clone(sys)
end_slice = b2.system.clone(sys)

sys.add(f)
start_slice.add(l)
end_slice.add(m)

#start point calculations
start_points_sys = b2.system.clone(sys)
start_points_sys.add(l)
start_points = nag_algorithm.ZeroDim(start_points_sys, mptype='adaptive')

#homotopy setup
start_hom_sys = b2.system.concatenate(sys, start_slice)
end_hom_sys = b2.system.concatenate(sys, end_slice)
hom = nag_algorithm.moving_homotopy(sys, start_hom_sys, end_hom_sys)

#slice moving
target = b2.system.concatenate(sys, end_slice)
moved = nag_algorithm.user_homotopy(hom, start_points, target, endgame='cauchy') #<- error here
moved.solve()
results = moved.solutions()
```

### LINKED ISSUE #259: Rare nondeterministic GMP 'Cannot allocate memory' SIGABRT (uninitialized mpfr/mpc numpy slot; ADR-0006 residual)
## Summary

A **rare, nondeterministic** hard crash (SIGABRT) inside GMP can occur the first time a multiprecision solve / `to_dataframe()` touches `mpfr`/`mpc` values, when `numpy`, `pandas`, and `matplotlib.pyplot` are all imported **before** `import bertini`. It looks like a residual case of the uninitialized `mpfr`/`mpc` numpy-slot family addressed in ADR-0006 (`docs/adr/0006-eigenpy-uninitialized-numpy-slot-guards.md`).

## The crash

```
GNU MP: Cannot allocate memory (size=575897802350002184)
Aborted (core dumped)        # exit code 134 (SIGABRT)
```

The absurd allocation size is the tell: GMP read an **uninitialized pointer/limb count** as a length — i.e. it operated on an `mpfr`/`mpc` whose limbs were never initialized.

## When it appears

Observed once while building a plotting tutorial, in a script whose import order was:

```python
import warnings, matplotlib
matplotlib.use("Agg")
import numpy as np, pandas as pd
import matplotlib.pyplot as plt
import bertini as pb                      # <-- bertini imported AFTER numpy/pandas/matplotlib
from bertini.nag_algorithm import ZeroDim
# ... build a 2-variable system, solver.solve() (adaptive precision), solver.to_dataframe() ...
```

Correlation found while narrowing it down:

- `import bertini` **first**, then numpy/pandas/matplotlib → never crashed.
- Each of numpy / pandas / matplotlib.pyplot imported **alone** before bertini → did not crash.
- All three imported before bertini (as above) → crashed **once**, then **did not reproduce in ~35 subsequent runs of the identical script**.

So it is heap-state-dependent and very rare, not a deterministic import-order failure. The import order only seems to raise the probability (presumably by changing what garbage occupies the uninitialized slot).

## Reproduction

Not reliably reproducible — it fired roughly **1 in 35 runs** of the script above. A stress loop is the most likely way to catch it again:

```bash
for i in $(seq 1 200); do
  python that_script.py || { echo "CRASH on run $i"; break; }
done
```

## Likely cause

ADR-0006 added guards for uninitialized `mpfr`/`mpc` numpy slots in the eigenpy bridge. This appears to be a remaining path those guards don't cover — an `mpfr`/`mpc` numpy scalar/array element used before its limbs are initialized. Importing numpy/pandas/matplotlib first may shift heap contents so the uninitialized limb pointer occasionally points at a value GMP then treats as a gigantic length.

## Environment

- bertini2: current `develop` (3.0.0.dev line), locally built `_pybertini`
- Python 3.14.5, Linux aarch64 (glibc 2.41)
- numpy 2.4.6, pandas 3.0.3, matplotlib 3.10.9

## Severity

Low frequency, but a hard SIGABRT with no Python-level traceback — and the triggering pattern (`import pandas`/`numpy` before `import bertini`) is completely ordinary user code. Worth chasing the residual uninitialized-slot path so the ADR-0006 guarantee is total.

### LINKED ISSUE #260: User homotopy crash when all system parameters contain the same function
When retrieving the solutions from a user homotopy where all system parameters of the user homotopy contain the same function, there is a crash with the following message:

set_prec.c:32: MPFR assertion failed: ((p) >= 1 && (p) <= ((mpfr_prec_t) ((((mpfr_uprec_t) -1) >> 1) - 256)))
zsh: abort      ipython

```
import bertini as b2
from bertini import nag_algorithm

x,y = b2.Variable('x'), b2.Variable('y')
vars = b2.VariableGroup([x, y])
#Lemniscate of Gerono system
f = x**4 - x**2 - y**2

#intersection lines from l to m
l = 2 * x - 7 * y - 1
m = x - 3 * y

#system setup
sys = b2.System()
sys.add_variable_group(vars)

start_slice = b2.system.clone(sys)
end_slice = b2.system.clone(sys)

sys.add(f)
start_slice.add(l)
end_slice.add(m)

#start point calculations
start_points_sys = b2.system.clone(sys)
start_points_sys.add(l)
start_points = nag_algorithm.ZeroDim(start_points_sys, mptype='adaptive')
start_points.solve()
start_points = start_points.finite_solutions() #<- crash here

#print("start points:\n",start_points)

#homotopy setup

hom = nag_algorithm.moving_homotopy(sys, start_slice, end_slice)

print("hom:\n",hom.describe())

#slice moving
target = b2.system.concatenate(sys, end_slice)
moved = nag_algorithm.user_homotopy(hom, start_points, target, endgame='cauchy')
moved.solve()
results = moved.finite_solutions() #<- crash here
```

### LINKED ISSUE #261: Parameters for the nag_algorithm.user_homotopy() are excessive
The parameters for the user_homotopy() require the homotopy and the start points, but also a target system. This system should already be included in the homotopy, so it makes little sense to require it again as its own parameter.

### LINKED ISSUE #263: Segfault when calling .function() on a system formed from a slice
Calling function() to retrieve a function from a system that is converted from a generic slice causes a segfault.

```
import bertini as b2
from bertini.nag_algorithm import Slice
x, y, z = b2.Variable('x'), b2.Variable('y'), b2.Variable('z')
vars = b2.VariableGroup([x,y,z])
s = Slice.random_complex(vars, 1)
sys = s.as_system()
print(sys.function(0)) #<- segfault here
```

### LINKED ISSUE #264: Printing a system converted from a generic slice
When printing a system converted from a generic slice, all functions appear as "c.[x, y, z, 1]" 

```
import bertini as b2
from bertini.nag_algorithm import Slice
x, y, z = b2.Variable('x'), b2.Variable('y'), b2.Variable('z')
vars = b2.VariableGroup([x,y,z])
s = Slice.random_complex(vars, 1)
sys = s.as_system()

print(s)
print(sys.describe())
```

ofloveandhate:
This PR merges the entire `ofloveandhate/b2` `develop` line into upstream — ~70 internal PRs. It's large, so this description indexes the work by theme, links each internal PR, and lists the upstream issues it closes. Going forward I'll open PRs directly against `bertiniteam/develop` so this roll-up isn't needed again.

## Highlights

- **~10× faster multiprecision** eval (tiered SLP arithmetic, allocation-free eval, CSE, in-place LU) and **5–7×** faster well-conditioned AMP solves by staying in double where safe.
- **Content-addressed function trees and systems** — hash-consing/interning, symbolic Jacobian, `Seal()`.
- **Parallel-by-default solving** without requiring MPI, plus MPI serialization & reproducibility fixes.
- **Friendlier Python surface**: flattened namespace, solutions in user coordinates, `to_dataframe`, sympy bridge, Unicode identifiers, revamped config model.
- The **`bertini2` CLI ships in the PyPI wheel** and emits Bertini-1.7-compatible solution files.
- **Durable, resumable output with provenance**: every solve writes a plain-text **structured output directory** — content-addressed inputs, one results file per run, walkable provenance chains — and consults it first, so a killed run *finishes* on rerun. Content-identity digests (systems, configs, seeds) make `seed=42` mean the exact same homotopy forever.
- Substantial **tracker/endgame** correctness (Cauchy security & pole-zone guards, predict/correct rewrite, adaptive-numeric-type AMP endgames).
- **Executable, verified docs** (doctest), a C++ doc-lint gate, and ADRs for load-bearing decisions.

## Upstream issues closed

- Closes #156 — pi computed inefficiently → ofloveandhate/b2#21
- Closes #239 — ship the (non-MPI) CLI in the wheel → ofloveandhate/b2#46
- Closes #253 — refresh solving-at-scale timing tables → ofloveandhate/b2#56
- Closes #255 — observer subsystem refactor → ofloveandhate/b2#23
- Closes #256 — Concatenate using cloned systems → ofloveandhate/b2#26
- Closes #258 — `ZeroDim…TotalDegree` not iterable → ofloveandhate/b2#30, #31, #33
- Closes #259 — nondeterministic GMP SIGABRT (uninitialized mpfr/mpc slot) → ofloveandhate/b2#32
- Closes #263 — segfault `.function()` on a slice-derived system → ofloveandhate/b2#51
- Closes #264 — printing a system converted from a generic slice → ofloveandhate/b2#52
- Closes #51 — system generator (classic input export) → ofloveandhate/b2#26
- Closes #74 — parser builds generic, not integer, power nodes → ofloveandhate/b2#40, #7
- Closes #87 — centralize/seed random number generation → ofloveandhate/b2#17
- Closes #123 — add `__str__` methods to many pybertini types → ofloveandhate/b2#7
- Closes #128 — expose the m-hom start system to pybertini → ofloveandhate/b2#18
- Closes #130 — tutorial on endgame usage → ofloveandhate/b2#22
- Closes #131 — tutorial: solve a system start-to-finish → ofloveandhate/b2#22
- Closes #133 — tutorial on user-defined homotopies → ofloveandhate/b2#22
- Closes #148 — pybertini complexes not pickleable → ofloveandhate/b2#21
- Closes #210 — build a System from a list of functions → ofloveandhate/b2#3, #65
- Closes #228 — stringification produces too many parentheses → ofloveandhate/b2#7
- Closes #232 — improve CI build time → ofloveandhate/b2#15
- Closes #235 — endgames need a genuinely adaptive precision mode → ofloveandhate/b2#55
- Closes #236 — make observers from Python with callbacks → ofloveandhate/b2#23
- Closes #243 — auto-randomize-and-filter overdetermined inputs → ofloveandhate/b2#49, #19
- Closes #247 — reparseable export (Bertini-1 classic format) → ofloveandhate/b2#26, #34, #39
- Closes #249 — rewrite predictor/corrector (per-track probe, pure kernel) → ofloveandhate/b2#43
- Closes #251 — non-mutating (copy-on-write) Simplify → ofloveandhate/b2#25, #7
- Closes #71 — TotalDegree constructor wrong for a non-affine variable group → ofloveandhate/b2#48
- Closes #95 — Reset wrapper for generic nodes (mooted by SLP-only evaluation) → ofloveandhate/b2#6
- Closes #100 — variable-values-random-by-default (mooted by SLP-only evaluation) → ofloveandhate/b2#25
- Closes #107 — optimize resetting of nodes in the function tree (mooted by SLP-only evaluation) → ofloveandhate/b2#25
- Closes #116 — mark leaf core types `final` (node hierarchy de-virtualized) → ofloveandhate/b2#6
- Closes #129 — move the start system out from under `system` → ofloveandhate/b2#48, #49
- Closes #132 — user homotopy missing its link to start points → ofloveandhate/b2#18, #49
- Closes #154 — precision consistency when homogenizing variables → ofloveandhate/b2#25, #26
- Closes #198 — computational inefficiency in predict/correct FullStep → ofloveandhate/b2#43
- Closes #199 — rename ODE predictor stages/variables → ofloveandhate/b2#43
- Closes #260 — user-homotopy crash (duplicate parameter functions) → ofloveandhate/b2#33
- Closes #261 — `user_homotopy()` parameters excessive → ofloveandhate/b2#49

### Multiprecision & evaluation performance
- ofloveandhate/b2#3 — grabbag of precision & step-size improvements (folds in the CLI/MPI/pytest work of fork #4).
- ofloveandhate/b2#5 — run well-conditioned AMP paths in double precision (5–7×).
- ofloveandhate/b2#40 — tiered SLP arithmetic + allocation-free mp eval (~10×).
- ofloveandhate/b2#41 — mimalloc + GMP allocator, AMP start-at-double, cheaper norms/temporaries.
- ofloveandhate/b2#44 — AMP Criterion B uses the Newton residual, not size-proportion (cyclic-5 11s → 3.5s).
- ofloveandhate/b2#58 — stop over-reading the SLP instruction tape by one word in `Eval`.
- ofloveandhate/b2#60 — SLP CSE: product power-fold + value-numbering (30–45% faster mp eval).
- ofloveandhate/b2#71 — stateful in-place multiprecision LU solver (replaces Eigen in the tracker).

Tiered SLP arithmetic roadmap

### Function tree & content-addressed systems
- ofloveandhate/b2#6 — eliminate diamond inheritance; de-virtualize the node hierarchy.
- ofloveandhate/b2#7 — function-tree introspection, precedence-aware printing, simplified differentiation, Hessian.
- ofloveandhate/b2#8 — fix `float()`/`complex()` on multiprec scalars (numpy-recursion segfault).
- ofloveandhate/b2#25 — hash-consing the function tree: interning, canonical variables & ordering, functional `Simplified()`.
- ofloveandhate/b2#37 — rename function-tree node `Float` → `Complex`.
- ofloveandhate/b2#38 — standardize scalar numeric type names (real/complex × dbl/mp).
- ofloveandhate/b2#59 — symbolic Jacobian in the core + `random_matrix` + nullvector criticality.
- ofloveandhate/b2#64 — hash-consing of Systems: content identity, `Seal()`, interning.

critical_points curve — two interlocking circles

### Trackers, endgames & adaptive precision
- ofloveandhate/b2#42 — fix inverted Cauchy `SecurityLevel` guard (truncate diverging paths).
- ofloveandhate/b2#43 — collapse predict/correct to a `StepMetadata` API + shared per-path probe.
- ofloveandhate/b2#50 — Cauchy c/k probe determinism, start-system sanity tests, per-path timing.
- ofloveandhate/b2#55 — adaptive-numeric-type AMP endgames + well-conditioned random coefficients.
- ofloveandhate/b2#69 — fully initialize the default `AMPConfig` (deterministic epsilon/Phi/Psi).
- ofloveandhate/b2#70 — Cauchy pole-component operating-zone check — never `Success` at a non-root.

AMP endgame session work
AMP endgame numeric-type flowchart

### Solve pipeline, start systems & randomization
- ofloveandhate/b2#12 — solutions in user coordinates by default; internal coords explicit; `homogenize_point` lift (ADR-0013).
- ofloveandhate/b2#17 — unify `RandomMp` onto the shared `ThreadEngine` so one seed controls all types.
- ofloveandhate/b2#18 — block-composed System: fold the polynomial path into blocks; MHom + linear algebra on blocks; expose m-hom start system.
- ofloveandhate/b2#19 — randomize overdetermined systems via a first-class `RandomizationBlock`.
- ofloveandhate/b2#48 — de-template ZeroDim on the start system; add linear-product TotalDegree.
- ofloveandhate/b2#49 — split ZeroDim into `HomotopySolver` (engine) + `ZeroDimSolver` (algorithm); auto square-up & filter.

### Parallelism
- ofloveandhate/b2#21 — ZeroDim crossed-path resolution, speculative full-path parallelism, reproducibility, pickling & MPI-portability.
- ofloveandhate/b2#36 — MPI-less shared-memory threading: parallel-by-default solves.
- ofloveandhate/b2#56 — fix parallel (MPI) multihomogeneous solve serialization; refresh scaling tables; restore `%` comment parsing.

### Witness sets, slices & NID enablement
- ofloveandhate/b2#34 — expose `WitnessSet` + a block-backed `Slice` (enable NID); classic slice/witness emission.
- ofloveandhate/b2#51 — `function()` on slice-derived systems + `system.slices()`.
- ofloveandhate/b2#52 — show low-precision coefficients in system `describe`.

### Python API & user experience
- ofloveandhate/b2#13 — sympy bridge: exact two-way conversion + solve round trip.
- ofloveandhate/b2#28 — config UX: string-valued fields, `owner.update(**fields)` router, `slice_` rename, `ZeroDimConfig` de-template.
- ofloveandhate/b2#29 — `FixedPrecisionConfig.precision`.
- ofloveandhate/b2#30 — ZeroDim solution access: `all_solutions`, `infinite_solutions`, `to_dataframe`.
- ofloveandhate/b2#32 — `to_dataframe` keeps the whole solution in one column.
- ofloveandhate/b2#62 — Unicode + emoji identifiers, name validation, collision-safe path variables.
- ofloveandhate/b2#65 — flatten the public Python API (breaking, pre-release); `System.add_functions`.
- ofloveandhate/b2#66 — config-model cleanup: `max_num_refinements` + single-source `final_tolerance`.

### CLI, distribution & versioning
- ofloveandhate/b2#39 — emit Bertini-1.7-compatible solution files + external-solver benchmark.
- ofloveandhate/b2#46 — ship the `bertini2` CLI in the PyPI wheel (`pip install bertini2` → `bertini2` on PATH).
- ofloveandhate/b2#61 — single-source `VERSION` file + full version in C++/CLI.
- (fork #4, folded into #3 — CLI exe, MPI parallelism, seeded RNG, pytest suite, benchmarks.)

### Bugfixes (cross-cutting)
- ofloveandhate/b2#14 — zero-dim coordinate-test tolerances respect the values' working precision.
- ofloveandhate/b2#26 — post-#25 grab-bag: precision fixes, Bertini-1 emitter, Concatenate fix, gamma conditioning, CI fallout.
- ofloveandhate/b2#31 — friendly error when a solver is passed to `user_homotopy`.
- ofloveandhate/b2#33 — reject equations in the wrong block of a moving homotopy.
- ofloveandhate/b2#47 — add pandas to test deps; document the numpy-reduction gotcha for mp dtypes.
- ofloveandhate/b2#53 — fix `moving_slice` `eval_time_derivative` doctest precision mismatch.

### Documentation, ADRs & CI
- ofloveandhate/b2#9 — single CI run per PR push; secret-gated GitLab sync for all forks.
- ofloveandhate/b2#10 — remove expression syntax from a workflow comment (it broke file validation).
- ofloveandhate/b2#11 — ADR-0010/0011/0012 for the function_tree overhaul; index + 0006 updates.
- ofloveandhate/b2#15 — ccache for bertini2 compiles (linux+macos wheels, unix C++ tests).
- ofloveandhate/b2#16 — explicit template instantiation of the endgame/ZeroDim universe (ADR-0014).
- ofloveandhate/b2#20, #22, #24 — documentation improvements; make tutorial code blocks executable & verified (doctest).
- ofloveandhate/b2#23 — observer subsystem refactor + Python path-visualization/callback observers.
- ofloveandhate/b2#27 — ADRs (gamma, fixed-multiple precision) + cyclic-5 cross-software regression test.
- ofloveandhate/b2#35 — resolve macOS compiler warnings.
- ofloveandhate/b2#54 — add a C++ doc-lint gate + fix rotted Doxygen docs.
- ofloveandhate/b2#63 — one-folder-per-tutorial, complete runnable examples, release-refresh mechanism.
- ofloveandhate/b2#67 — one-time deterministic regeneration of committed plot images.

### Structured output directory: records, resume, provenance
- ofloveandhate/b2#68 — every solve writes a durable, plain-text **structured output directory** and consults it first, so a crashed solve *finishes* on rerun (ensure-answered `solve()`: recorded paths recall instead of recompute). The directory is self-describing and needs no software to read.
  - **Content identity for reproducibility**: versioned canonical encodings + SHA-256 digests for Systems (`b2sysenc`), configs (`b2cfgenc`), and seed-rooted randomness (`b2rand`) — equal objects digest equally *forever*, across machines, compilers, and library versions; append-only version registries make an un-bumped encoding change fail loudly. `seed=42` names the exact homotopy — same gamma, patch, and start points — everywhere.
  - **Three truth stores**: `history/` (what was asked, when), `results/` (one append-only JSONL file per run — every computed path with its endpoint and metadata), `definitions/` (content-addressed inputs: systems, configs, givens); `README`/`INDEX` are derived, rebuildable views.
  - **Reproducible sessions**: effective per-solve seeds make every recorded run replayable standalone; "a solve leaves the session as it found it" (default precision + RNG streams restored on exit); `recall` events narrate re-asks; records are **on by default** for the CLI, `bertini.solve`, and the bare `ZeroDimSolver`/`HomotopySolver` classes.
  - **Casual surface**: `bertini.solve` / `save` / `load` / `annotate`; `Solution` = points that remember their provenance; chained solves thread provenance across runs, walkable back to canonical start labels or user-supplied givens; pandas + networkx navigation of the records.
  - **Cauchy endgame corrections** (found via the records' own `function_residual` and a cyclic-6 audit against Bertini 1, refining #70): the security check watches the extrapolated **endpoint** for divergence to infinity (not the loop samples) and truncates **only in the operating zone**; the pole-growth truncation is dropped in favor of the operating-zone acceptance gate alone — cyclic-6 recovers its full 156 nonsingular solutions.
  - ADR-0042..0047; record-format spec at `docs/records/b2rec-1.md`.


ofloveandhate:
EPIC.

ofloveandhate:
20260706_arc_state_diagram_structured_output_directory

ofloveandhate:
i am so sorry i didn't just do the work directly against bertiniteam.  a bad habit haunted me.  i learned my lesson.

PR #266, opened 2026-07-03
fix(trackers): fully initialize default AMPConfig (deterministic epsilon/Phi/Psi)

ofloveandhate:
## Problem

The `Test wheel on windows-3.13` job has been intermittently failing on:

```
pickle_test.py::test_config_struct_round_trips[AMPConfig]
```

with two objects that print **identically** yet compare unequal:

```
assert AMPConfig(..., epsilon=nan, phi=0.0, psi=2.121995841e-314, ...)
    == AMPConfig(..., epsilon=nan, phi=0.0, psi=2.121995841e-314, ...)
```

`psi=2.12e-314` is a denormal and `epsilon=nan` — i.e. **uninitialized memory**. Equality fails because the C++ `operator==` compares `epsilon`, and `nan != nan`. This is why the same commit passed on one CI run and failed on the next: it only bites when the garbage happens to land as NaN.

## Root cause

`AdaptiveMultiplePrecisionConfig`'s default constructor set `coefficient_bound`, `degree_bound`, and the safety-digit fields, but left **`epsilon`, `Phi`, `Psi` uninitialized**.

## Fix

These three bounds are always recomputed from the system before tracking (`AMPConfigFrom` / `SetAMPConfigFrom`), so the defaults are placeholders — but a public default-constructed config still has to be deterministic. Initialize `epsilon` to the single-variable value (`1`) and derive `Phi`/`Psi` from the default bounds via the existing `SetPhiPsiFromBounds()`, so the object is fully initialized and internally consistent:

```
epsilon = 1, Phi = 20000 (= D(D-1)B), Psi = 5000 (= D·B)
```

## Verification

- `pickle_test.py` 17/17, full `pytest` suite green (bar the pre-existing editable-build CLI-binary tests, unrelated).
- `bash tools/doclint.sh` clean (new `///` doc comment on the ctor).
- Bindings recompile clean.


ofloveandhate:
Retargeting: this fix should land on ofloveandhate:develop (PR #238's head) so it flows into #238 for free. Reopening against the fork's develop.

Issue #304, opened 2026-07-08
add `is_distinct_up_to`


ofloveandhate:
because i often want to know whether two points are the same, up to some tolerance.  the norm is the infinity norm.

ofloveandhate:
Addressed in #305 — `bertini.is_distinct_up_to(p, q, tol)`: True iff the points differ by more than `tol` in the infinity norm. Accepts `real_mp` / `complex_mp` / `real_dbl` / `complex_dbl` vectors, so you don't have to think about which number type you're holding. Backed by a reusable C++ `IsDistinct` util.

Issue #303, opened 2026-07-08
do i need a HomogeneousVariableGroup type?


ofloveandhate:
to help distinguish?  or are they all just `VariableGroup`s?

Issue #302, opened 2026-07-08
solution_metadata should have a way to get the metadata, given just a point


ofloveandhate:
so i can get the solutions, and then get a solutions, and then get just the metadata for that point

ofloveandhate:
Addressed in #305 — `solver.metadata_for(point, tol=...)` returns the cluster representative's `SolutionMetaData` (one record, carrying `.multiplicity`); `coincident=True` returns the list of every coincident copy. The return type depends on the call, never on the point's multiplicity.

Issue #301, opened 2026-07-08
better help using abs, round etc, from numpy


ofloveandhate:
for real_mp and complex_mp types

i want to be able to type 

```
np.round(complex_mp)
```

or 

````
np.abs(complex_mp)
```

and it just works

ofloveandhate:
Addressed in #305 via helpers — `bertini.abs` / `round` / `sum` / `norm` / `real` / `imag` / `conj` / `is_real` over scalar / list / numpy array, staying mp-native (`round` uses Decimal so it keeps arbitrary precision). **Note:** overlaps with the native ufunc work on `feature/numpy_compat` — reconcile before closing.

PR #305, opened 2026-07-08
feat: UI quality-of-life batch (#293–#304) + solution group projection — 3.1.0.dev1

### LINKED ISSUE #293: TypeError when getting the variable group of a system.
When trying to access the variable group of the system, it returns what I think is the memory address of the variables and not a list of the variables.

TypeError: No registered converter was able to produce a C++ rvalue of type std::__1::shared_ptr<bertini::node::Variable> from this Python object of type VariableGroup

```
import bertini as b2
from bertini import nag_algorithm
from bertini.nag_algorithm import Slice

x, y, z = b2.Variable('x'), b2.Variable('y'), b2.Variable('z')
f1 = (y - x**2) * (x**2 + y**2 + z**2 - 1) * (x - 2)
f2 = (z - x**3) * (x**2 + y**2 + z**2 - 1) * (y - 2)
f3 = (z - x**3) * (y - x**2) * (x**2 + y**2 + z**2 - 1) * (z - 2)

sys = b2.System()
vars = b2.VariableGroup([x,y,z])
sys.add_variable_group(vars)
sys.add_functions([f1,f2,f3])

slice = Slice.random_complex(sys.variable_groups(), len(sys.degrees()) - 1) #<- Error here
```

ofloveandhate:
## UI quality-of-life batch (issues #293–#304) + solution group projection

These rough edges surfaced while implementing the numerical real cellular decomposition algorithm in
Python. The through-line goal: get from a polynomial system to the projection values of critical
points **as `real_mp`, straight through numpy, without a code maze**.

Per the "put reusable primitives in core" doctrine, anything with a C++ home landed in **core +
bindings** (behind C++ tests); only presentation/interop sugar stayed pure-Python.

### C++ core + bindings
- **#294** — `Slice::RandomReal`'s orthogonal path was silently using the *complex* conjugate-orthonormal
  matrix, so a "real" slice came out complex; now genuinely real. New `random_real` / `random_complex` /
  `random_vector` factories (continuous, seed-reproducible — the generic-direction tool, unlike the
  quantized orthonormal `random_matrix`).
- **#304** — reusable `IsDistinct` / `IsSamePoint` infinity-norm util → `bertini.is_distinct_up_to`
  (accepts `real_mp` / `complex_mp` / `real_dbl` / `complex_dbl` points).
- **#299** — `merge_multiplicities` (default **True**) on the solver's own `SolutionsWhere` + finite
  accessors — dedup where the multiplicity clustering lives.
- **#302** — `metadata_for(point)`: returns the cluster **representative** by default (one record,
  carrying `.multiplicity`), or `coincident=True` for the list of every copy — return type never
  depends on the point's multiplicity.
- **#297** — `System.functions()` / `copy_functions()`.
- **#300** — `node.eval` accepts a point array or a `{Variable-or-name: value}` dict (not just kwargs),
  still returning `complex_mp` natively.
- **Group projection** — `System.coordinates_of(point, group)` + a `group=` keyword on every solution
  getter and `to_dataframe`: project solutions onto one variable group (projective groups returned
  as-is). Reusable for the eventual C++ decomposition.

### Python layer
- **#293** — `variables(['x','y','z'])`; `add_variable_group` accepts loose variables / a list / a single
  Variable / a VariableGroup; `Slice.random_*` gives a clear message (and unwraps a one-group sequence)
  instead of an opaque converter error.
- **#296** — `sys.clone()` method.
- **#295** — `bertini.sympy_bridge` exposed lazily + nodes auto-sympify (`_sympy_`), so
  `sympy.sympify(node)` / `sympy.Matrix(array)` / `sympy.det(J)` just work.
- **#294 sugar** — `VariableGroup @ coeffs` → the linear-combination node.

### #298 / #301 (numpy interop) — note the overlap
Adds mp-native elementwise helpers `bertini.real/imag/abs/conj/round/sum/norm/is_real` (the
numpy↔eigenpy dtype boundary is unpatchable for identity-seeded reductions; documented in
`known_gotchas`). **These likely overlap with the native ufunc coverage on `feature/numpy_compat`** —
left for the maintainer to reconcile (the helpers still stand alone for `is_real` / dedup / norm).

### Deferred
- **#303** (HomogeneousVariableGroup) — out of scope by decision.

## Verification
- C++: `test_classes` + `test_nag_algorithms` green (new regression tests for every C++ change).
- Python: **660 passed, 4 skipped** (the one initial failure was the intended #299 default change; that
  test was updated).
- Acceptance artifact: a polished marimo notebook (in the sibling `curve_decomposition_in_python` repo,
  original left untouched) runs end-to-end — projection values arrive as `real_mp`, plus a 3D plot of
  the real critical points.

## Heads-up for review
- **Version bump 3.0.0rc1 → 3.1.0.dev1** is included (this branch *is* the 3.1.0.dev1 work). Drop the
  first commit if you'd rather bump separately.
- **The projective-nullvector reformulation does not track** — mathematically cleaner (only critical
  points survive) but the mhom homotopy fails every path for this bilinear structure. The notebook uses
  the working affine+patch formulation and documents the projective attempt as an open question for the
  start-system / endgame work.
- **`int(real_mp)` segfaults** (pre-existing `real_mp` binding bug; `float`/`str` fine) — hit
  incidentally, out of scope here, worth a follow-up (`real_mp.__int__`).
- `doclint` not run locally (no doxygen in this env); every new public C++ entity carries a matching
  Doxygen comment, so the CI gate should pass.

Closes #293, #294, #295, #296, #297, #299, #300, #302, #304.
(#298 and #301 addressed via helpers — see the overlap note above before closing.)


ofloveandhate:
20260708_ui_polish_architecture

ofloveandhate:
there's some competing work in here, with #306 .  will deal with in a minute

PR #306, opened 2026-07-08
numpy compat: full ufunc coverage, sorting, safe reductions, tolerance comparisons for mp dtypes

### LINKED ISSUE #298: using `.real` in numpy does NOT get the real part
Image

### LINKED ISSUE #301: better help using abs, round etc, from numpy

ofloveandhate:
## What this PR delivers

The numpy gotchas for the `real_mp` / `complex_mp` dtypes are gone. The scary "Known gotchas" page is retired; its replacement, `docs/source/numpy.rst` ("Multiprecision numbers and NumPy"), documents the interop as the feature it now is. Full decision record in **ADR-0051**; includes the post-#305 reconciliation.

### Full ufunc coverage
`np.abs`, `np.conj`, `np.exp`/`log`/`log2`/`log10`/`expm1`/`log1p`, all trig/hyperbolic + inverses, `power`, `sign`, `reciprocal`, `minimum`/`maximum`/`fmin`/`fmax`, `floor`/`ceil`/`trunc`/`rint`, `isnan`/`isinf`/`isfinite`/`signbit`, `arctan2`, `hypot`, `copysign`, `mod`/`fmod`/`floor_divide` — previously all `TypeError: ufunc not supported`. Every loop reads through `value_or_zero` (ADR-0006 doctrine) and calls the same boost::multiprecision function the `multiprec` scalar functions bind, so `np.f(a)[i] == mp.f(a[i])` exactly. numpy-semantics corners done deliberately: `rint` is half-to-even via direct `mpfr_rint` (boost's is half-away; registered for complex too, component-wise, so `np.round` works on complex arrays), `mod` takes the divisor's sign, `minimum`/`maximum` propagate nan while `fmin`/`fmax` ignore it.

### Sorting & reductions — and the bug hiding under them
`HardenCompare`/`HardenArgMinMax` fill the dtype slots for `real_mp`: `np.sort`, `np.argsort`, `np.searchsorted`, `np.median`, `np.argmax`/`argmin`, `np.max`/`min` (complex stays unordered by design). `np.sum`/`np.prod`/`np.mean` work (verified numpy 2.3.2 & 2.4.6, regression-tested) — but investigating them exposed the real hazard: **`getitem` returned `boost::ref` into the numpy buffer** (stock eigenpy behavior), so a scalar extracted from a temporary array dangled — `np.mean` returned a silently wrong `0`, and `s = np.sum(v)` kept past the statement SIGABRT'd in `str()`. This was the root of the ADR-0031 / #259 aliasing class. `getitem` now returns an **owned copy**, killing the class at the source.

### The float64 boundary: closed for values, open for tolerance comparisons
`double→mp` casts stay **unsafe** — floats must not pollute polynomial-system construction. But tolerance *orderings* against a double are registered (`np.abs(a - b) < 1e-10`, both operand orders): a comparison yields a bool, no float value enters an mp computation, and the compare is exact (1e-22 is not lost against a 1e-30 tolerance). Mirrors the C++ solvers' double `ToleranceT` and the scalar `GreatLessVisitor<T,double>` precedent. Mixed equality and arithmetic stay blocked; `np.isclose`/`allclose` still raise (they *compute* with float64 tolerances). `arr.astype(float)` / `astype(complex)` are the explicit conscious truncations (`mp→complex128` cast registered unsafe; it never had been). Registration-order note: casts must register **before** the ufunc loops, or numpy permanently ignores them.

### Component access on complex arrays — "a crash is better than incorrect values"
numpy's `ndarray.real`/`.imag` are C getsets gated on `PyArray_ISCOMPLEX`, a hardwired builtin-type check with no user-dtype hook (verified against numpy 2.4.x `getset.c`; neither the legacy API nor NEP 42 offers one) — on mp-complex arrays they return silently wrong values. Defended everywhere reachable:
- **`Solution` overrides `.real`/`.imag`** at the subclass level — solve results are simply correct;
- **`bertini._numpy_guard`**, installed at import, wraps `np.real`/`np.imag`/`np.angle` to raise a `TypeError` naming the right tool on plain mp-complex input (pass-through otherwise; `angle` raises for every mp-complex input since it branches on dtype);
- sanctioned array accessors `bertini.real/imag/arg` and `mp.real/imag/arg` (en route: fixed the scalar `mp.imag`, which had returned the **real** part since it was written — copy-paste bug);
- the raw attributes on a plain self-built ndarray are the one unreachable spelling — documented, pinned by test, and tracked upstream-ward in #307.

### One namespace for the whole vocabulary
`bertini.operators` is now polymorphic: `from bertini.operators import *` gives `sin`/`cos`/`exp`/... that dispatch per argument (symbolic node for a Variable/expression; numeric through the native precision-preserving loops for mp scalars, numpy containers, lists, and plain numbers), plus `abs`/`arg`/`real`/`imag`/`conj`/`round`/`sum`/`norm`/`is_real` and `E`/`Pi`/`I`. The top-level elementary functions are rebound to the polymorphic versions (a strict superset), and `bertini.arg` + the hyperbolics join the top level. No more remembering that `arg` lives in `bertini.multiprec` while `imag` is top-level.

### #305 reconciliation
Develop (the UI QoL batch) is merged in. Its interim `bertini.real/imag/abs/conj/sum/norm/is_real` helpers keep their API but now ride the native ufunc loops on mp-dtype arrays, falling back to element-wise work for lists. Two latent #305 bugs fixed: the helpers' scalar fallbacks called bare `abs()`/`round()`, which resolve to the module's own shadowing names → infinite recursion on plain python input; and the docs claim that the numpy boundary "cannot be patched in the bindings" (this PR is the patch).

## Tests / verification
- `numpy_ufuncs_test.py` (~120) + `operators_test.py` (17): exact agreement with the scalar functions, semantics corners, precision preservation through every loop shape (including the mixed real/complex division path — loops re-tag via `at_precision_of`), unwritten-slot safety per ADR-0006, sorting incl. nan-wins argmax, reductions, tolerance comparisons, astype, guard behavior, polymorphic dispatch, and named regressions for the dangling-scalar, `mp.imag`, and shadowed-builtin bugs.
- Full suite **793 passed**; **169 sphinx doctests** green (the numpy page is executable); doclint clean; end-to-end verified on a precision-40 solve with numpy ops on the returned solutions.

## Linked issues
- Closes #298 — `.real` in numpy doesn't get the real part: Solution overrides + guarded numpy functions + `bertini.real/imag`/`mp.real/imag/arg` accessors; the scalar `mp.imag` bug found under it is fixed. The unreachable plain-attribute residue is #307.
- Closes #301 — `np.abs(complex_mp)` / `np.round(complex_mp)` and the rest of the ufunc family now just work.
- #307 (filed from this work) tracks the numpy-side shortcoming and the watch-items for future numpy versions.


ofloveandhate:
2026-07-08_numpy_compat_ufunc_coverage

ofloveandhate:
Fable make many hours of work, very simple.  I love using this system.  Let me focus on the math!!!

ofloveandhate:
and a while later, i understand that numpy has a significant BUG for complex user types: imag and real will always return incorrect values.  what?!?!?!?  #307 is the lasting memory of it, and I should probably do the work to help numpy fix this.  for now, users will get an exception if they try to, because i would rather have an error, than an incorrect value/result.

ofloveandhate:
here's the final state of the PR's work, in picture form (assuming i get green on CI and don't have to do more work):

2026-07-08_numpy_compat_ufunc_coverage

PR #331, opened 2026-07-12
feat(results): ZeroDimResult answer type; SolveResult wraps it; positive-dim guard

ofloveandhate:
The `SolveResult`-as-records-wrapper keystone, and the first member of the result taxonomy.

## `ZeroDimResult` — the typed answer
The answer of a zero-dimensional solve: the distinct finite solutions plus category views, a portable snapshot detached from the solver.
- `finite` — the finite solutions (the deliverable; distinct, one per multiplicity cluster, each carrying its records provenance)
- `real` / `singular` / `nonsingular` — the finite solutions by class
- `at_infinity` — diverged endpoints; `nonsolutions` — finite endpoints that aren't solutions of the target
- iterable / `len` / indexable over the finite solutions; exposed as `bertini.ZeroDimResult`

Both solver families produce one — `ZeroDimSolver.solve()` **and** `HomotopySolver.solve()` return a `SolveResult(answer=ZeroDimResult)` (the name is about the *answer's* dimensionality — isolated points — not the solver).

## `SolveResult` is now a records decorator
A thin wrapper around `.answer` (a `ZeroDimResult` today; an `NIDResult` once NID lands) that adds the run id / directory / recall count. `solutions` / `__iter__` / `__len__` / `__getitem__` delegate to the answer, so it still behaves exactly like the list of solutions it was — no caller change.

## Positive-dimensional guard
`bertini.solve` on an under-determined (positive-dimensional) system now raises a clear `NotImplementedError` — its solution set is a variety, not isolated points, so it needs numerical irreducible decomposition (which returns an `NIDResult` of witness sets, the other taxonomy member) — instead of surfacing the raw solver `RuntimeError`. `NIDResult` (wrapping the already-bound `WitnessSet`) slots into `.answer` when NID itself is implemented.

Full Python suite green (records + feasibility shown: 34 passed, 1 skipped; whole suite 880).

### DIFF
--- a/python/bertini/__init__.py
+++ b/python/bertini/__init__.py
@@ -117,7 +117,7 @@ from ._coefficients import coefficient, coefficients
 # the casual records surface: solve / save / load over the structured output directory
 from .records import (solve, save, load, annotate, solutions_of, provenance,
                       recording, records_dir, runs, tracks, provenance_graph,
-                      plot_chain, Solution, SolveResult)
+                      plot_chain, Solution, SolveResult, ZeroDimResult)
 from . import records
 
 # attach the friendly system-building methods and Slice.from_coefficients (was bertini.linalg.*)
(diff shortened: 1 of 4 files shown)


ofloveandhate:
i love how much faster CI is since I worked on it a few days ago.  It had been a bit painful.  Now its like 30 minutes from a warm cache.

PR #345, opened 2026-07-15
fix(trackers): condition-number refresh counter passed by value made frequency_of_CN_estimation inert

ofloveandhate:
## What

`ExplicitRKPredictor::SetNormsCond` took its refresh counter `num_steps_since_last_condition_number_computation` **by value**. The tracker initializes that counter *to* `frequency_of_CN_estimation`, so the member never advanced past that value, `counter >= frequency` held on every step, and the condition-number estimate silently refreshed **every step** — the frequency knob was inert.

- At the default `frequency_of_CN_estimation = 1`: behavior identical (why nothing ever noticed).
- For any larger value: the skip the user asked for never happened (one extra probe triangular-solve per step), and the counter-reset semantics in the function body were dead code.

## Change

One-character fix (`unsigned` → `unsigned &`) plus the matching explicit-instantiation signatures, and a doc-comment guardrail explaining why the parameter must be by reference.

## Test

`condition_number_refresh_honors_frequency` (tracking_basics): a `SuccessfulStep` observer records `LatestConditionNumber()` along a path whose conditioning varies smoothly (`y² − t − 1`), with frequency 3. The estimate must refresh at least once but change at most every third step. Verified red with the bug restored, green with the fix; full `test_tracking_basics` suite green; doclint green.

Found during a paper-vs-implementation audit of the AMP tracker against BHSW (the audit found the tracker otherwise faithful to the published algorithm).

### DIFF
--- a/core/include/bertini2/trackers/explicit_predictors.hpp
+++ b/core/include/bertini2/trackers/explicit_predictors.hpp
@@ -636,10 +636,10 @@ namespace bertini{
 				/// \param[out] norm_J Set to ||J||.
 				/// \param[out] norm_J_inverse Set to the estimate of ||J^{-1}||.
 				/// \param[out] condition_number_estimate Set to the product of the two norms.
-				/// \param num_steps_since_last_condition_number_computation Steps elapsed since the last estimate.
+				/// \param[in,out] num_steps_since_last_condition_number_computation Steps elapsed since the last estimate; advanced/reset here, so it MUST be the caller's counter by reference (passing by value silently turns any frequency > 1 into "never refresh").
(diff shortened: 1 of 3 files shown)


ofloveandhate:
The macOS failure was a flaw in the regression test, not the fix: the test used a univariate system, and for a 1×1 Jacobian the condition estimate is constant **by algebra** (`‖J‖·‖J⁻¹r‖ = |r|`, the fixed probe's norm) — so the 'it does refresh' assertion could only pass via floating-point jitter, which macOS's libm doesn't produce. Replaced with a 2-variable system whose conditioning genuinely varies; re-verified red-with-bug (estimate changed on every transition, 10/10) / green-with-fix (changes ≤ every 3rd step) locally.

PR #353, opened 2026-07-16
fix(pseg): cycle-number bound is a ceiling; Hermite interpolation evaluates the actual interpolant

ofloveandhate:
Two power-series endgame defects found by auditing PSEG against the Cauchy fix history ("many of the bugs fixed in cauchy probably have analogues in PSEG"). Both are old; both were masked by tests whose expectations were calibrated to the buggy implementations.

## 1. `max_cycle_number` was a floor, not a ceiling

`ComputeBoundOnCycleNumber` applied the config with `max()` instead of `min()`, so "largest cycle number to consider" never bounded the candidate search. A near-unity sample ratio — slow convergence: high multiplicity, or a slow diverger — made the amplified estimate itself the bound (hundreds of Hermite solves per approximation), and a ratio within ~1e-14 of 1 amplifies past `UINT_MAX`, where the unclamped conversion to `unsigned` is undefined behavior.

Now clamped in the real type before any conversion; `!(amplified < ceiling)` also routes inf/NaN to the ceiling. The old test literally pinned the inverted behavior ("max_cycle_num implemented max(5,6) = 6"); a named regression test covers both the slow-ratio cap and the UB-magnitude clamp.

## 2. `HermiteInterpolateAndSolve` did not evaluate the Hermite interpolant

The Horner reconstruction walked the doubled node list at half speed (node `z_ii` paired with coefficient `a_{2·ii}`), evaluating a polynomial that is **not** the interpolant: a cubic — which 3-node Hermite must reproduce exactly — came back 1.6e-5 off. The limit as the sample window slides to the target was unaffected, so PSEG still converged; but at degraded order, costing extra iterations per path and delivering poorer approximations than the samples support.

The pinned expectation ("found using matlab") was itself not the Hermite value, and its loose tolerance let the wrong code pass. All expectations re-derived in exact rational arithmetic:

- the interpolant of the `x^8+1` test data at 0 is exactly `0.999999998837890625` (pinned at 1e-20);
- the window-halving errors are `1.162e-9 → 4.539e-12 → 1.773e-14` (~256× per halving — the restored convergence order, visible in-test);
- new named regression test pins cubic exactness at 1e-20.

## Records note

Fix 2 changes the values PSEG produces — this is a *computational semantics* change in the sense of the records-identity doctrine (the behavior-epoch case). The epochs mechanism doesn't exist yet; this note is the marker.

## Tests

387 endgame cases green (all three PSEG flavors × precisions), full ctest battery green, Python suite 926 passed / 3 skipped, both doclints clean.

### DIFF
--- a/core/include/bertini2/endgames/interpolation.hpp
+++ b/core/include/bertini2/endgames/interpolation.hpp
@@ -112,20 +112,19 @@ template<typename ComplexT>
 		}
 	}
 
-	//Start of Result from Hermite polynomial, this is using the diagonal of the 
-	//finite difference matrix.
-	Vec<ComplexT> Result = space_differences(2*num_sample_points - 1,2*num_sample_points - 1); 
-
-
-	//This builds the hermite polynomial from the highest term down. 
-	//As we multiply the previous result we will construct the highest term down to the last term.
-	for (unsigned ii=num_sample_points-1; ii >= 1; --ii)
-	{
-		Result = ((Result*(target_time - time_differences(ii)) + space_differences(2*ii, 2*ii)) * (target_time - time_differences(ii-1)) + space_differences(2*ii-1, 2*ii-1)).eval();  
-	}
-	
-	// Last term in hermite polynomial.
-	return (Result * (target_time - time_differences(0)) + space_differences(0,0)).eval(); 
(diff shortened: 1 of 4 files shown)


ofloveandhate:
I rest easier, with this Hermite interpolant bug solved.

PR #374, opened 2026-07-25
Standalone NewtonRefine: the deflated-system sharpening primitive (core + bindings + python)

### LINKED ISSUE #373: Expose tracker Refine (core Newton) to Python
## Motivation

The pure-Python cellular decomposition work has established that decomposition reliability hinges on the accuracy of a small population of boundary-stage points (critical points above all): a crit point accurate to eps yields slices whose tangency geometry is perturbed by ~sqrt(eps), which materializes as duplicate-vertex mis-assembly downstream. The fix is Newton sharpening of these points at registration time — and the core already owns the right primitive, so the Python layer should not re-implement Newton.

## What exists in core

`trackers/base_tracker.hpp` provides tracker-owned Newton refinement, in fixed and adaptive precision:

```cpp
SuccessCode Refine(Vec<C>& new_space, Vec<C> const& start_point, C const& current_time);  // implicit tolerance
SuccessCode Refine(Vec<C>& new_space, Vec<C> const& start_point, C const& current_time,
                   RealT const& tolerance, unsigned max_iterations);                       // explicit
```

The explicit-tolerance overload is exactly the sharpening primitive: refine this point, on this tracker's system, at this time, to this tolerance. Neither overload is currently exposed in `python_bindings/src/tracker_export.cpp`.

(`nag_algorithms/sharpen.hpp` is an empty scaffold — the algorithm-level Sharpen can come later; this issue is only about exposing the existing tracker primitive.)

## Proposed binding shape

- **Return the refined vector; do not bind the output-reference signature.** The natural `Vec<C>&` out-param sits adjacent to a `C const&` time scalar — the exact eigenpy writable-Ref/adjacent-scalar corruption pattern of ADR-0001. A small wrapper that allocates, calls `Refine`, and returns `(SuccessCode, Vec)` sidesteps the hazard entirely; scalar time passed **by value** per the ADR.
- Bind both overloads (implicit and explicit tolerance) on the fixed-precision and AMP tracker classes.
- Tolerance as `real` (mp), max_iterations as unsigned; defaults mirroring `NewtonConfig`.

## Acceptance

- `tracker.refine(point, time)` and `tracker.refine(point, time, tolerance=..., max_iterations=...)` from Python on both tracker types, returning the refined point plus the SuccessCode.
- A Python test refining a loosely-tracked nonsingular solution to a tight tolerance and verifying the residual/step-size claim.
- No re-implementation of Newton in the pure-Python layer.

## Context

Needed by the decomposition-reliability work: sharpening all boundary-stage points (crit/preimage/singular/sphere) at the vertex-registration boundary so identity decisions and projection-value authorities run on honest, tight error bars. Nonsingular points use this primitive directly on the already-squared (randomized) tracking systems; singular points will layer isosingular deflation on top of the same call.

ofloveandhate:
## Standalone NewtonRefine: tracker-free Newton against a (deflated) System

The sharpening primitive the cellular-decomposition reliability work needs: refine a point against a **square autonomous system** (or one with a path variable, at a supplied time) to a requested consecutive-step tolerance, at the precision of the inputs — with no tracker construction or configuration.

The tracker-owned `Refine` is already exposed to Python, but it refines against *the tracker's own system*. Sharpening singular points requires Newton on the point's **deflation** — a plain autonomous `System` no tracker is configured around — which is exactly what this free function serves. Overdetermined systems are refused with instructions to square by randomization, matching production practice. Closes #373 (with a correction to that issue's premise: `tracker.refine` was already exposed; the standalone is the genuinely missing piece).

### What's here

- `core/include/bertini2/nag_algorithms/newton_refine.hpp` — `bertini::algorithm::NewtonRefine`, returning `(code, point, achieved step norm, iterations)`. Fully doxygen'd.
- Bindings: `newton_refine` free function (both arities), following ADR-0001 (refined vector returned by value; scalar time by value).
- `bertini.newton_refine` pure-Python wrapper (numpydoc; py_doclint at baseline).

### Tests — the isosingular hierarchy, in miniature

C++ (`core/test/nag_algorithms/newton_refine.cpp`), each sub-millisecond:

1. **nonsingular root**: quadratic convergence to 1e-40 in ≤10 iterations
2. **overdetermined refusal**: friendly `std::runtime_error` with squaring instructions
3. **plain double root**: provably *stalls* (linear halving cannot reach 1e-40 in 25 iterations) — the disease, isolated
4. **double cone singularity**: one deflation (`[f; ∇f]`, hand-randomized with small-prime matrices whose nonsingularity at the target is verified in comments) → lands ON the origin to 1e-40
5. **whitney umbrella, handle point (0,0,1)**: a smooth point of the singular curve — deflation + pinning slice, randomized → ON the point
6. **whitney umbrella, pinch point**: a singular point OF the singular embedded curve — the stage-1 deflation is *itself singular there and provably fails*; stage 2 (appending 2×2 minors of the stage-1 Jacobian) succeeds → ON the origin

Python interface tests mirror 3 of these end-to-end (cone from `bertini.newton_refine`: 4 iterations, 1e-6 offset → <1e-40 from the singularity); full Python suite green (932 passed).

### The one bug the tests caught

`linalg::PartialPivLU` is stateful and requires `ChangeSize` before `Factor`/`Solve`; without it both run over a 0-dimensional system and "succeed" without computing. Caught by test 3 *demanding failure* and getting a phantom success — skeptical tests earning their keep.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -239,6 +239,7 @@ set(io_parsing_settings_headers
 
 set(nag_algorithms_headers
 	include/bertini2/nag_algorithms/midpath_check.hpp
+	include/bertini2/nag_algorithms/newton_refine.hpp
 	include/bertini2/nag_algorithms/numerical_irreducible_decomposition.hpp
 	include/bertini2/nag_algorithms/output.hpp
 	include/bertini2/nag_algorithms/sharpen.hpp
@@ -931,6 +932,7 @@ if (ENABLE_UNIT_TESTING)
 		test/nag_algorithms/zero_dim_records.cpp
 		test/nag_algorithms/numerical_irreducible_decomposition.cpp
 		test/nag_algorithms/trace.cpp
+		test/nag_algorithms/newton_refine.cpp
 	)
 
 	add_executable(test_nag_algorithms ${B2_NAG_ALGORITHMS_TEST})
--- a/core/include/bertini2/nag_algorithms/newton_refine.hpp
+++ b/core/include/bertini2/nag_algorithms/newton_refine.hpp
@@ -1,0 +1,149 @@
+//This file is part of Bertini 2.
+//
+//bertini2/nag_algorithms/newton_refine.hpp is free software: you can redistribute it and/or modify
+//it under the terms of the GNU General Public License as published by
+//the Free Software Foundation, either version 3 of the License, or
+//(at your option) any later version.
+//
+//bertini2/nag_algorithms/newton_refine.hpp is distributed in the hope that it will be useful,
+//but WITHOUT ANY WARRANTY; without even the implied warranty of
+//MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//GNU General Public License for more details.
+//
+//You should have received a copy of the GNU General Public License
+//along with bertini2/nag_algorithms/newton_refine.hpp.  If not, see <http://www.gnu.org/licenses/>.
+//
+// silviana amethyst, university of wisconsin eau claire
+
+/**
+\file bertini2/nag_algorithms/newton_refine.hpp
+
+\brief Standalone Newton refinement of a point against a System -- no tracker required.
+
+The sharpening primitive: given a square system and an approximate solution, iterate
+Newton's method until consecutive approximations agree to a requested tolerance.  The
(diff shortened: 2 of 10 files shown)


ofloveandhate:
it's really satisfying to be able to refine to a singular point using deflation.  this PR does not add deflation, but it does add refinement in a better way.

PR #375, opened 2026-07-25
Classic-format I/O hardening: tolerant parsing, full-precision printing, NeverStarted named

ofloveandhate:
Four small fixes to the classic-format I/O layer and one binding name, all surfaced by the cellular-decomposition reliability work driving b2 as an engine:

- **`parse.system` tolerates a CONFIG section and `INPUT`/`END;` separators** — real-world Bertini-1 input files carry them; the parser now skips the config block and separators instead of failing.
- **`parse.system` reads `(re,im)` complex coefficient literals** — the classic complex spelling parses to exact coefficients.
- **float coefficients print at FULL precision in classic input** (`Complex::print` via `str(0)`) — previously ostream defaults truncated to ~6 digits, so a System round-tripped through classic format silently lost its coefficients (caught when saved decomposition bundles disagreed with the systems that produced them).
- **`SuccessCode::NeverStarted` gets its name in the bindings** — the anonymous `SuccessCode(-1)` that meant "the endgame never ran because the pre-endgame track failed"; naming it enables honest `pre:<cause>` failure attribution downstream.

### DIFF
--- a/core/src/function_tree/symbols/number.cpp
+++ b/core/src/function_tree/symbols/number.cpp
@@ -61,11 +61,15 @@ void Integer::print(std::ostream & target) const
 void Complex::print(std::ostream & target) const
 {
 	// real-valued floats print bare; the complex pair form is reserved for
-	// genuinely complex values
+	// genuinely complex values.  str(0) prints EVERY digit the stored value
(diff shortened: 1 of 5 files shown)


ofloveandhate (commented):
solid

PR #387, opened 2026-07-28
System.eval accepts any point precision; mixed time/space precisions throw (closes #377)

### LINKED ISSUE #377: Precision alignment is a pervasive footgun: evaluation should not throw when precisions differ
## The smell

Working with mixed-precision objects requires every caller to manually align three separate precision states -- the point's, the System's, and the System's SLP's -- or evaluation throws:

    RuntimeError: variable_values and SLP must be of same precision.  respective precisions: 60 120
    RuntimeError: precision of input point in SetVariables (60) must match the precision of the system (30)

Today's field evidence, from wiring Newton sharpening (elevated-precision refinement of critical points) into the cellular decomposer: introducing points at ambient+30 digits surfaced **seven distinct downstream sites** that crashed on precision mismatch (rank checks, slice residuals, moving-homotopy start prep, crossing solves, records serialization/reload), each needing its own hand-written `target = max(point, system, ambient); system.precision(target); lift(point, target)` boilerplate.  The consuming codebase has grown a small ecosystem of coping helpers (`lift_precision`, `_eval_at_common_precision`, per-site alignment) and STILL got bitten by:

- **the precision ratchet**: `system.precision(...)` persists on shared/cached systems, so repeated elevated-precision operations compound (60 -> 90 -> 120) until unrelated eval sites disagree;
- **asymmetric helpers**: an 'up-only' lift silently keeps higher precision when a caller asks for ambient, poisoning downstream consumers;
- **ultimately a two-tier delivery workaround**: compute sharp at elevated precision, hand results back DOWN-cast to ambient, because delivering full precision would require touching every consumer.

This is poor UI: the caller's intent is always simply "evaluate f at x", and the library makes that intent expressible only through state surgery on a shared mutable object.

## Related existing sharp edges (same family)

- mixed-precision arithmetic mis-tagging results (real_mp/complex_mp division tagging precision 0);
- `default_precision()` vs literal/coefficient precision divergence after AMP solves;
- SLP precision diverging from its System's reported precision after solves at other precisions.

## The ask

A coherent, documented precision model with evaluation that just works:

1. `System::Eval(x)` (and Jacobian) should ALIGN rather than throw -- evaluate at the max of the operand precisions (re-tagging the SLP internally), or at an explicitly requested precision.  Throwing on mismatch turns every precision boundary into a crash site.
2. Precision changes on a System should not be load-bearing global state for callers: either make `precision()` cheap and automatic at eval time, or provide a scoped/temporary form so shared systems cannot be left ratcheted.
3. One documented ownership rule for who holds precision truth (point, System, SLP, DefaultPrecision) and how they reconcile -- the current three-way split with hard failure on disagreement is the root of the whole class.

This aligns with the longer-term direction already recorded for AMP work (accuracy-aware rematerialization rather than incidental `.precision` calls scattered through algorithms).

ofloveandhate:
Closes #377.

## The footgun

`System::SetVariables` threw whenever a multiprecision point's precision differed from the system's internal precision. Since the ambient default precision drifts (most notably after an adaptive-precision solve), every consumer of `System::Eval` eventually hit

```
precision of input point in SetVariables (40) must match the precision of the system (30).
```

at a call site that did nothing wrong. The cellular/NID layers had grown repeated align-then-eval workarounds; each new eval call site re-discovered the crash.

## The contract now

Evaluation is stateless (blocks are value-in; the SLP is the evaluator), so **the input defines the working precision**:

- `SetVariables` changes the system's internal precision to match the incoming point — in both directions. A caller evaluating at precision *p* gets the evaluation at precision *p*; the system's previous precision is not their concern. Precision-managing callers (the trackers) always match already and take the fast path unchanged.
- **Time and variables are different in kind**: both are supplied in the *same* call, so a precision mismatch between them is caller incoherence, not internal-state drift. `EvalInPlace` / `JacobianInPlace` (the funnels for every point+time evaluation) now throw with a message naming both precisions and instructing the caller to align them. Previously `SetPathVariable` checked nothing, so mixed time slipped in silently.
- `BERTINI_DISABLE_PRECISION_CHECKS` builds behave as before (no checks, no alignment).

## Tests

- C++ (`test_classes`): `system_evaluate_mpfr_any_point_precision` — a precision-30 system follows a 50-digit point up (result precision 50, values exact) and a 20-digit point down; Jacobian too. `system_evaluate_mixed_time_precision_throws` — mismatched time throws for both `Eval` and `JacobianInPlace`; aligned inputs evaluate at their common precision.
- Python (`python/test/classes/system_test.py`): the eval surface accepts any point precision, and the mixed time/space message reaches Python as a `RuntimeError` naming both precisions and the word "align".

Both doc-lint passes clean (C++ and Python).

### DIFF
--- a/core/include/bertini2/system/system.hpp
+++ b/core/include/bertini2/system/system.hpp
@@ -333,9 +333,9 @@ namespace bertini {
 		/**
 		 Evaluate the system, provided a path variable is defined for the system, in place.
 
-		 \throws std::runtime_error, if a path variable is NOT defined, and you passed it a value.  Also throws if the number of variables doesn't match.
+		 \throws std::runtime_error, if a path variable is NOT defined, and you passed it a value.  Also throws if the number of variables doesn't match, or if (for multiprecision types) the precisions of ``variable_values`` and ``path_variable_value`` differ -- both are supplied in the same call, so align them before evaluating.
 		 \tparam T the number-type for return.  Probably complex_dbl=std::complex<double>, or complex_mp=bertini::complex_mp.
-		 
+
 		 \param function_values The vector to write the function values into.
 		 \param variable_values The values of the variables, for the evaluation.
 		 \param path_variable_value The current value of the path variable.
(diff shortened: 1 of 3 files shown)


ofloveandhate:
Closing in favour of #394 and #399, which fix #377 by making evaluation follow the precision of the point it is given and then removing System::precision altogether (ADR-0057). This branch changed System::SetVariables instead; the two approaches conflict, and the 3.5 work has been running on #394/#399.

PR #388, opened 2026-07-30
slice mover, regen base level Jupyter notebook drafts

ElderStack:
Adds the drafts of SliceMover and RegenCascade that I've been working over the summer. The RegenCascade file is incomplete. It generates codim 1 startpoints but I was unable to get codim 2+ regeneration working.


ofloveandhate:
Thank you.  I'll review.

ofloveandhate (commented):
Work towards an implementation of numerical irreducible decomposition.   A foundation for the full algorithm.  Slice mover is more complete than the regenerative cascade.  Neither ready for direct inclusion in the library, but both will be useful as I continue the implementation.  Thanks!

ofloveandhate:
This will come in as code in a new `drafts` folder, which will eventually disappear as the real implementation is completed.

PR #393, opened 2026-08-27
fix(endgames): PSEG left previous_approximation_ a copy of final; bind the approximation accessors

ofloveandhate:
## Why

Deflation needs to decide the rank of a Jacobian at an *approximate* root, and a single
spectrum at a single point cannot do it — a tiny singular value is indistinguishable from a
perturbation artifact. What separates them is watching how the value **moves** as the
approximation improves. The endgame already computes a second, deliberately coarser sample of
the root and the distance to it, and then throws both away: `PreviousApproximation()` and
`ApproximateError()` exist in C++ but have never been reachable from Python.

Binding them turned up a bug that made one of them useless anyway.

## The bug

The power series endgame assigned `previous_approximation_ = final_approximation_` at the
**bottom** of its convergence loop while testing the loop condition at the **top**, so the
assignment ran one last time on the way out. After every successful PSEG run the two vectors
were **equal**. Cauchy never had this — it returns from its acceptance gate before the
corresponding assignment — so the two endgames disagreed about their own post-run state.

Fixed in both PSEG drivers (fixed-precision `RunImpl` and adaptive `RunImplAMP`) by breaking
before the assignment, matching Cauchy's shape. The loop's own entry condition is deliberately
kept so a `FinalTolerance() >= 1` still refuses to enter at all, exactly as before.

**This reached further than the accessor.** `ZeroDimSolver` computes
`accuracy_estimate_user_coords` as the distance between the final approximation and the
previous one:

```cpp
smd.accuracy_estimate_user_coords =
    (DehomogenizePoint(solutions_post_endgame_[i]) -
     DehomogenizePoint(ctx.endgame.PreviousApproximation<BaseComplexT>())).lpNorm<Infinity>();
```

For power-series solves that was **identically zero** — an exactly-perfect accuracy reported
for every such path. The value flows into the records archive
(`records/solver_recording.hpp`) and MPI `path_result`. It is not part of any digest, so the
persistent-identity contract is untouched.

## Also fixed

`EndgameBase::approximate_error_` was never initialized, so `ApproximateError()` read an
indeterminate value before any run — not acceptable for a member now exposed to Python.

It is initialized to **infinity, deliberately not NaN**. NaN is the more honest "no estimate",
but it loses *every* relational comparison, and the gates read it in both directions:
`cauchy.hpp` has `approximate_error_ < FinalTolerance()` (NaN → false → not converged, fine),
but `powerseries.hpp` has `while (approximate_error_ > FinalTolerance())` — NaN there is false,
the loop never runs, and the endgame reports instant success on whatever happened to be in
`final_approximation_`. Infinity fails the `<` gate and passes the `>` gate, and is not a false
claim: with no approximation computed, the honest error bound is unbounded.

## What is bound

`previous_approximation()` and `approximate_error()` on every endgame flavor.
`final_approximation()` was already bound. Returned by value like the existing accessor, since
the endgame overwrites its approximation vectors on the next run (ADR-0051's owned-copy
doctrine for eigenpy-backed vectors).

Working end to end:

```
PSEG    pre-run error = inf
        final    = 1.000000000000804
        previous = 1.000000000004163      <- a genuine predecessor, not a copy
        approximate_error = 3.359091e-12   ||final-previous|| = 3.359091e-12
CAUCHY  final = 1,  previous = 1+1.54e-18j
        approximate_error = 2.220500e-16   ||final-previous|| = 2.220500e-16
```

## Tests

`approximation_accessors_are_a_coherent_triple`, added to both generic endgame test headers so
it runs for every tracker flavor at every ambient precision, plus Python interface tests.

Two things the tests deliberately do **not** assert, both learned by measurement:

- **Not** `ApproximateError() > 0`. `fixed_multiple_cauchy` at precision 16 converges with two
  successive approximations agreeing *bitwise* on this system, whose root is exactly 1.
  Coherence is the invariant; a nonzero gap is not.
- **Not** an anchored comparison. `diff <= 1e-10 * max(1, err)` **passes against the unfixed
  endgame** — with the bug `gap = 0` and `err ~ 1e-12`, so it reduces to `1e-12 <= 1e-10`. The
  converged error is below any absolute tolerance one would think to write, so the check is
  relative to the error itself, where the bug is a ratio of exactly 1.

Verified the regression test actually bites: reverting only the PSEG change and rebuilding
gives **8 targeted failures** across every PSEG flavor and precision.

## Verification

- `ctest --test-dir build/core` — **10/10 suites pass** (428 endgame cases)
- `pytest python/test/` — **926 passed, 1 skipped**
- `python tools/py_doclint.py` — clean, at baseline
- C++ doc-lint not run locally (no doxygen on this machine); the diff adds no new public entity
  under `core/include`, so the zero-undocumented ratchet is unaffected.

## ADR

Assessed, and I do not think one is warranted. The one thing a reader might "helpfully" undo is
the loop now carrying both a `while (error > tolerance)` condition and an
`if (error <= tolerance) break;`, which looks redundant — but that is a bug fix carried by a
regression test that fails loudly across 8 suites, which the ADR bar explicitly exempts. The
reasoning for both the break and the infinity sentinel is written inline at each site. Happy to
add one if you would rather have the guardrail.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,38 @@ _______________________________________________________________________________
 
 _______________________________________________________________________________
 
+## [3.5.0] - unreleased
+
+### Added
+
+- Python bindings for two endgame accessors that already existed in C++ but were unreachable:
+  `previous_approximation()` and `approximate_error()`, alongside the already-bound
+  `final_approximation()`.  Together they give the pair of successive root approximations the
+  endgame's own convergence test compares, plus the infinity norm between them -- a second
+  sample of the root at a known, coarser accuracy, which is what lets a caller judge how a
+  derived quantity (the singular values of a Jacobian, say) behaves as the approximation
+  improves, rather than thresholding it at one point.
+
+### Fixed
+
+- The power series endgame left `previous_approximation_` holding a COPY of
+  `final_approximation_` after every successful run.  It assigned the two at the bottom of its
+  convergence loop while testing the loop condition at the top, so the assignment ran one final
+  time on the way out.  The Cauchy endgame never had this -- it returns from its acceptance gate
+  before the corresponding assignment -- so the two endgames disagreed about their own post-run
+  state.  Power series now matches Cauchy.  Consequences: `PreviousApproximation()` is now a
+  genuine predecessor for both endgames, and `ZeroDimSolver`'s reported
+  `accuracy_estimate_user_coords` -- computed as the distance between the final approximation and
+  the previous one -- is no longer identically zero for power-series solves, which had it
+  reporting an exactly-perfect accuracy for every such path.
+- `EndgameBase::approximate_error_` was left uninitialized, so `ApproximateError()` read an
+  indeterminate value before any run.  Now initialized to infinity, which is the only safe
+  sentinel: the convergence gates compare it in both directions, and NaN -- which loses every
+  relational comparison -- would make the power series loop's `error > tolerance` test false and
+  skip the loop entirely, reporting instant success.
+
+_______________________________________________________________________________
+
 ## [3.4.0] - 2026-07-16
 
 A one-call way to build a linear slice that passes through a chosen point, an endgame-hardening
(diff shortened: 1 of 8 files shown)


ofloveandhate (commented):
yep, mathematically correct, fixes a bug i made a decade ago.  THanks!

Issue #397, opened 2026-08-27
PowerOperator::Degree(VariableGroup) sums per-variable degrees instead of the total degree — inflates DegreeBound(), which feeds AMP


ofloveandhate:
## The bug

`PowerOperator::Degree(VariableGroup const&)` computes the total degree by **summing the
per-variable degrees**:

```cpp
auto multideg = MultiDegree(vars);
auto deg = 0;
std::for_each(multideg.begin(), multideg.end(), [&](int n){
        if (n < 0) deg = -1; else deg += n; });
return deg;
```

Summing is valid only for a **monomial**. For `(x+y)^2` the degree in `x` is 2 and the
degree in `y` is 2, but the total degree is **2**, not 4. The overcount scales with the
number of variables appearing in the base:

| expression | true degree | reported |
|---|---|---|
| `x^4` | 4 | 4 |
| `(x+y)^2` | 2 | **4** |
| `(x^2+y^2)^2` | 4 | **8** |
| `(x^2+y^2+z^2+3)^2` | 4 | **12** |

## Why it stayed hidden

The same expression takes two different paths. An integer exponent written in C++ or Python
builds an **`IntegerPowerOperator`**, whose `Degree(VariableGroup)` is
`base_deg * exponent` and has always been right. The **classic-input parser** builds the
generic **`PowerOperator`**, which has this bug.

Both print identically, and both give the correct *ungrouped* `Degrees()`. Only
`Degrees(Variables())` differs — and that is exactly what `System::DegreeBound()` calls.

```python
S = System with f0 = (x+y)^2 built via Python's **
P = bertini.parse.system("variable_group x, y;\nfunction f0;\nf0 = (x+y)^2;\n")

str(S) == str(P)                  # both print  f_0 = (x+y)^2
list(S.degrees()) == list(P.degrees())     # both [2]
list(S.degrees(vg)), list(P.degrees(vg))   # [2] vs [4]     <-- here
```

## Why it matters

`DegreeBound()` feeds **AMP**. So a system round-tripped through classic input tracks with a
different degree bound than the one it was built from — silently, with no error, in the
direction of demanding more precision than necessary. Any records/replay layer that stores
systems as classic-input text therefore replays them under different adaptive-precision
decisions than the original run, which undermines reproducibility.

## Fix

Give `Degree(VariableGroup)` the same shape as `PowerOperator::Degree(std::shared_ptr<Variable>)`
directly above it, which is already correct: analyse the exponent, and use the base's
[...]

Issue #396, opened 2026-08-27
to_classic_input() / parse.system() do not round-trip, and the mismatch fails SILENTLY (empty system, no error)


ofloveandhate:
## `to_classic_input()` and `parse.system()` look like inverses, and are not

```python
import bertini
from bertini.symbolics import Variable
x, y, z = Variable('x'), Variable('y'), Variable('z')
S = bertini.System(); vg = bertini.VariableGroup()
for v in (x, y, z): vg.append(v)
S.add_variable_group(vg)
S.add_functions([(x**2+y**2+z**2+3)**2 - 16*(x**2+y**2)])

bertini.parse.system(S.to_classic_input())
# RuntimeError: [SystemParser] parse error at line 2, col 1:
#   expected: "="   found: "tracktype: 0; mptype: 2; ..."
```

`to_classic_input()` emits a complete Bertini 1 file — `CONFIG ... END;` then
`INPUT ... END;` — which is exactly what it documents and what it is *for* (running the same
problem in Bertini 1 for cross-validation, added in c78def48). `parse.system()` accepts only
the bare INPUT-section body. So the two are not inverses, and nothing says so.

## The part that makes it dangerous: the failure is SILENT

Strip the CONFIG block and it gets worse rather than better:

```python
bertini.parse.system("INPUT\nvariable_group x, y, z;\nfunction f0;\nf0 = x^2+y^2+z^2-1;\nEND;\n")
#  -> returns a System.  num_functions() == 0, num_variables() == 0.  No error.

bertini.parse.system("variable_group x, y, z;\nfunction f0;\nf0 = x^2+y^2+z^2-1;\n")
#  -> num_functions() == 1, num_variables() == 3.   correct
```

Given the `INPUT ... END;` wrapper, the parser consumes **nothing**, returns an **empty
system**, and reports **no error**. A caller that round-trips through text gets an object
that is structurally valid and completely empty. Handing that empty system back to
`to_classic_input()` then segfaults the interpreter (#395), which is how this was found.

## The inverse already exists in C++ — it just isn't bound

`core/include/bertini2/io/classic_writer.hpp`:

```cpp
/// \brief The INPUT-section body as a string.  \see EmitSystem.
inline std::string SystemToClassic(System const& sys)          // <- the actual inverse
inline std::string SystemToClassicFile(System const& sys, ...)  // <- bound as to_classic_input()
```

`SystemToClassic` emits the body with "No `INPUT`/`END;` wrapper", which is precisely what
[...]

PR #398, opened 2026-08-27
fix(classic-io): three defects in the classic-input round trip (segfault, silent empty parse, inflated degree bound)

### LINKED ISSUE #395: System::CoefficientBound calls Eigen maxCoeff() on empty arrays: bertini.System().to_classic_input() segfaults
## Three lines

```python
import bertini
bertini.System().to_classic_input()      # Segmentation fault (core dumped)
```

Also crashes with variables present, as long as there are no functions:

```python
S = bertini.System(); vg = bertini.VariableGroup(); vg.append(Variable('x'))
S.add_variable_group(vg)
S.to_classic_input()                     # Segmentation fault
```

The trigger is **zero functions**. A system with functions but no variable group is fine —
it raises a clean `RuntimeError: SLP compile: a function references the variable 'x', which
is not in the system's variable ordering`, which is the behaviour one wants.

## Cause

`System::CoefficientBound<NumT>()` (`core/src/system/system.cpp:944`) ends with

```cpp
bound = max(f_vals.array().abs().maxCoeff(),
            dh_dx.array().abs().maxCoeff(), bound);
```

With no functions, `f_vals` is an empty vector and `dh_dx` an empty matrix, and Eigen's
`maxCoeff()` on an empty array is **undefined behaviour** — it asserts in a debug build and
reads out of bounds in a release one.

`DegreeBound()`, immediately below it in the same file, already guards precisely this case:

```cpp
auto degs = Degrees(Variables());
if (degs.empty())
    return 0;   // a system with no functions has no degree bound
```

`CoefficientBound` needs the same guard.

The Python path in is `to_classic_input()` → `EmitConfig` (`io/classic_writer.hpp`), which
emits `coefficientbound:` and so calls `CoefficientBound<complex_dbl>()`. But the bug is in
`CoefficientBound` itself, so any other caller reaching it with a function-less system
crashes too.

## Expected

A function-less system is a legitimate object — `bertini.System()` constructs one, and
`num_functions()`, `num_variables()` and `degrees()` all answer for it correctly (`0`, `0`,
`[]`). Its coefficient bound should be `0`, matching `DegreeBound()`'s convention, and no
call reachable from Python should ever be able to segfault.

## Found via

A downstream records layer that writes systems as classic-input text; a parse that silently
produced an empty system (filed separately) then handed that empty system back to
`to_classic_input()`, which crashed the interpreter rather than reporting anything.

Version: `3.5.0.dev0` (develop @ 8f0cd0f1), Linux, Python 3.14.

ofloveandhate:
Fixes #395, #396 and #397.

Pulling on "a records layer cannot reload the systems it wrote" turned up three
independent bugs, each of which hid the next.

## #395 — segfault

`System::CoefficientBound` ends in

```cpp
max(f_vals.array().abs().maxCoeff(), dh_dx.array().abs().maxCoeff(), bound)
```

and with no functions both arrays are **empty**, where Eigen's `maxCoeff()` is undefined
behaviour. `bertini.System().to_classic_input()` reached it — the CONFIG section emits
`coefficientbound:` — and crashed the interpreter.

`DegreeBound()`, four lines below in the same file, already guarded exactly this case and
returns 0. `CoefficientBound` never did. Now guarded on **both** dimensions, since `dh_dx`
is functions × variables and so is empty when either is zero.

## #396 — the round trip failed, and failed silently

`to_classic_input()` emits a complete Bertini 1 file (`CONFIG ... END;` then
`INPUT ... END;`) because its purpose is running the same problem in Bertini 1; the grammar
reads the INPUT-section **body**. They were never inverses.

Worse, the Python binding **discarded the parse result**:

```cpp
bertini::parsing::classic::parse(str.begin(), str.end(), res);
return res;              // parse() said false and never touched res
```

`parse()` answers `false` when the grammar matched nothing or did not consume the input, and
leaves the result untouched — so Python received a structurally valid, entirely **empty**
System and no error. A caller round-tripping through text carried on with zero functions.

Two fixes. `StripClassicFileWrappers()` (in `qi_files.hpp`) lets both `System::System(string)`
and the binding accept a full classic file — a **no-op** for text that is already a bare
body, so nothing that worked before changes. And the binding now raises, including when the
input "successfully" declares nothing at all (an empty string, or an `INPUT` section with an
empty body): matching zero declarations is a successful parse of the grammar, but it still
leaves the caller holding a System they did not ask for.

## #397 — and the one that actually bites

`PowerOperator::Degree(VariableGroup)` computed the total degree by **summing the
per-variable degrees**, which is valid only for a monomial. `(x+y)^2` has degree 2 in `x`
and 2 in `y` but **total degree 2**, and it reported 4. The overcount scales with the number
of variables in the base:

| expression | true degree | reported |
|---|---|---|
| `x^4` | 4 | 4 |
| `(x+y)^2` | 2 | **4** |
| `(x^2+y^2)^2` | 4 | **8** |
| `(x^2+y^2+z^2+3)^2` | 4 | **12** |

It stayed invisible because the same expression takes two paths: an integer exponent written
in C++ or Python builds an `IntegerPowerOperator`, whose group degree is `base_deg*exponent`
and always was right, while the classic parser builds this generic `PowerOperator`. Both
print identically, and both give the correct **ungrouped** `Degrees()`. Only
`Degrees(Variables())` differed — which is precisely what `System::DegreeBound()` calls, and
**DegreeBound feeds AMP**.

So a system round-tripped through classic input tracked under a *different
adaptive-precision regime* than the one it was built from, silently. Any records/replay
layer storing systems as classic text was replaying them under different precision decisions
than the original run.

Now the same shape as `PowerOperator::Degree(v)` directly above it, with the base's group
degree in place of its degree in one variable.

## Tests

Both languages, each case naming its issue: the empty-system crash; the full-file round
trip; the bare-body convention still parsing; unparseable and empty input raising instead of
yielding an empty system; the total degree of a power of a sum; and `DegreeBound()` agreeing
between a built and a parsed system.

One deliberate asymmetry worth flagging to a reviewer: the round-trip test compares the
**INPUT sections only**. `coefficientbound` is estimated by evaluating at *random* points, so
two emissions of the very same system legitimately differ there — `to_classic_input()` is not
byte-deterministic, and asserting that it is would be wrong.

Verified: `ctest` 10/10 suites, `pytest` 942 passed / 1 skipped.

## ADR

None proposed. Each fix restores documented or obvious intent rather than making a new
tradeoff, and all three are carried by regression tests that fail loudly if undone.

### DIFF
--- a/core/include/bertini2/io/parsing/qi_files.hpp
+++ b/core/include/bertini2/io/parsing/qi_files.hpp
@@ -57,6 +57,8 @@
 #include <iostream>
 #include <sstream>
 #include <stdexcept>
+#include <cctype>
+#include <cstring>
 #include <string>
 
 #include "bertini2/io/parsing/unicode_ident.hpp"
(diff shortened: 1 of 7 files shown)


ofloveandhate (commented):
Solid.  Tests prove some erroneous behaviour I had implemented my own self, is now gone.  I do with the parsing was in Boost.Spirit rather than plain string work, but it's simple enough that I'll accept it.

PR #399, opened 2026-08-27
refactor(precision)!: a System no longer carries a precision at all

ofloveandhate:
Closes the rest of #377 (asks 2 and 3). **Stacked on #394** — review that first; this branch
contains its commit, and the removal only makes sense once evaluation aligns itself.

## A System no longer carries a precision

The setter, the getter, and the `precision_` member are all gone, in C++ and in Python.

Once evaluation aligns to the point it is handed (#394), a System's precision had no job
left. What it did have was a cost — the boilerplate it forced on every caller:

```cpp
auto target = max(Precision(point), system.precision(), DefaultPrecision());
system.precision(target);
```

That is the complaint #377 opens with, and refusing to evaluate on a mismatch could wedge a
System outright with no escape through either setter. Two independent places had even grown
the same post-deserialize workaround, `sys.precision(sys.precision())`, to force a
re-materialize — `witness_set.hpp` and `system_export.cpp`. Both are gone.

## Where the state actually belongs

The objects that **hold** multiprecision values — each block's working coefficients, the
patch, the SLP's memory — already kept their own "materialized at" tag. The System's copy was
a duplicate of theirs.

Only the patch **elided** on it. The five blocks gained the same early return, and that is
what makes it affordable for `SetVariables` to fan out on *every* evaluation: the unchanged
path is now a handful of integer compares across seven holders.

The elision is deliberately **per holder and not at the System**, which is a soundness point
rather than a style one. A System-level early-out would need a System precision to compare
against, which no longer exists — and inventing one would skip work that is genuinely needed,
because holders can legitimately disagree. An SLP compiled at the ambient default while the
rest sat elsewhere *is* the desync of #377; a System-level cache would say "already at 30,
nothing to do" and skip the very repair. Each holder knows its own truth.

Result precision is likewise **derived**: `CoerceBlockOutputPrecision` takes its target from
the staged point instead of a stored field.

## Nothing remains to prepare

There is no "prepare"/"materialize at" entry point either. **Every evaluable type
self-aligns** to the precision of the point it is handed, under one name and one body:

```cpp
template <typename T>
void SyncPrecision(Vec<T> const& vars) const   // no-op for double
{
    if (vars.size() && Precision(vars(0)) != precision_)
        Precision(Precision(vars(0)));
}
```

called first thing in each `EvalInPlace` / `JacobianInPlace` / `TimeDerivInPlace`. Six types
have it: the four blocks, the patch, and the SLP.

**This is not a new invention.** `blend_block` and `randomization_block` already did exactly
this independently — the change makes the pattern *and the name* uniform. The patch tells the
same story from the other side: it carried a **commented-out assert** demanding that callers
match its precision, disabled because it could not be honoured. It now aligns itself instead.

An earlier draft of this branch kept a prepare-style method, on the theory that evaluations
consuming *staged* values (`EvalInPlace`/`JacobianInPlace` with no input point, as Newton
uses) need the system prepared beforehand. **That theory is false**: every such evaluation is
preceded by an `Eval(point)` that stages the point and aligns as a side effect. Removing all
**22** of its callers — 13 in the trackers/endgames/`zero_dim_solve`, 5 pushing into
sub-Systems from blocks, 4 in start systems — changed no test result. It was pure ceremony,
and a long self-documenting name made it look deliberate.

## Coefficient generation is a construct-time concern

The start systems (`mhom`, `total_degree_linear_product`) build their blocks' exact masters
under `DefaultPrecision(MaxPrecisionAllowed())` — which is all that matters, since the ambient
default governs the precision new mp values are *born* at. They no longer touch any working
precision. Setting the precision of working coefficients is an evaluation concern only.

## Invariants that ceased to exist were removed, not worked around

- `FixedPrecisionTracker::PrecisionSanityCheck` no longer asks the System what precision it
  is at.
- `euler_test`'s `assert(sys.precision()==50)` is gone — there is nothing left to report.
- The tracker observers' log formatting takes its digit count from the **values being
  printed**, not from the ambient default (which need not be the precision anything was
  computed at) and not from the System.

## Breaking

- **Python**: `System.precision` no longer exists, in either form.
- **Archive format**: `precision_` is no longer serialized. It was transient evaluation state
  — the same `serialize` already excludes `current_variable_values_` for exactly that reason —
  but a System archived by an older build will not load into a newer one. Boost archives carry
  Systems between MPI ranks within one run; durable storage is the records/JSON path
  (ADR-0042) and is unaffected.

## Notes for review

Most of the call-site diff is **deletion**. The three-line incantation

```cpp
DefaultPrecision(50);  sys.precision(50);  expanded.precision(50);
```

collapses to its first line, because the points carry 50 digits and the systems follow.

One test needed rethinking rather than converting. `system_identity_test` proved that two
Python handles are one C++ System by setting a precision through one and reading it back
through the other. With no precision to observe, it now stages a point through one handle and
evaluates with **no arguments** through the other — which only works if they share their
staged values. That is a better proof: it exercises the real shared transient state instead of
a bookkeeping field.

## Considered and rejected

Moving the staged values into a separate evaluation handle (`sys.At(point).Values()`).
Recorded in ADR-0057 so it is not re-proposed: it would tax the common case —
`sys.Eval(point)`, one call, no ceremony — to buy thread-safety that per-thread cloning
already provides. Direct evaluation is the interface worth protecting. `current_variable_values_`
stays: unlike `precision_` it is invisible to callers and does real work, backing the
no-argument `Eval()`/`Jacobian()` so the SLP's tape runs once and the Jacobian can be read
later without re-running it.

## ADR

**ADR-0057** — the ownership rule #377 asks for, with a table of who holds precision truth
(the point is the authority; every `precision_` is a derived cache tag; the exact masters are
the source of truth for values, so everything is *re-materialized, never re-drawn*), the
archive-format consequence, and the rejected handle design.

Verified: `ctest` 10/10 suites, `pytest` 934 passed / 1 skipped, both doclints clean —
including the AMP-heavy tracking, endgame and nag_algorithms suites, which are the ones that
would notice.


ofloveandhate:
omg finally.  this was long overdue and resolves a bunch of tension in the library.  i do note that if a time and space variables are used in eval, they still should be in the same precision, so "use any precision you want" is not allowed, in a strict sense.  but now at least the system is not a piece of a puzzle to resist the assembler.  it''s 14:28, and i'm listening to the album Musca by Matthew Herbert.  Human in the loop!!!

ofloveandhate (commented):
omg this is such a good PR.  it finally resolves this stupid problem i caused.

there are some places in this PR where comments explain things that used to be there, which will in time become useless.  at some point it doesn't matter what *was* there, only what is.  But i'll still accept this.

Issue #410, opened 2026-09-11
A path can fail steps forever: max_num_steps counts only SUCCESSFUL steps, and min_step_size is never reached


ofloveandhate:
## Summary

A path can enter a **provably non-terminating loop** in `AMPTracker`: it fails steps
forever, at ~6000 failures/second, while every guard that should stop it is structurally
unable to fire. Observed live: **3.9 million failed steps against 343 successful ones**, 30+
minutes at 100% CPU with no progress, and the enclosing `Solve()` never returns.

This is the concrete mechanism behind #400 ("bound what a single path may cost"). #400 asks
for a budget; this issue is that the *existing* guards are unreachable, so a budget is not
merely nice to have — right now nothing can stop this path at all.

## The three guards, and why none of them fires

```cpp
// trackers/amp_tracker.hpp:617, PreIterationCheck()
if (num_successful_steps_taken_ >= Get<Stepping>().max_num_steps)
    return SuccessCode::MaxNumStepsTaken;
if (current_stepsize_ < Get<Stepping>().min_step_size)
    return SuccessCode::MinStepSizeReached;
if (current_precision_ > Get<PrecConf>().maximum_precision)
    return SuccessCode::MaxPrecisionReached;
```

**1. The step budget counts only SUCCESSFUL steps.** `num_successful_steps_taken_` against
`max_num_steps` (default 100000). The observed path has 343 successes and 3.9M failures, so
it is at 0.3% of its budget and will never approach it. A failing path is effectively
un-budgeted.

**2. The stepsize never reaches `min_step_size`, so `MinStepSizeReached` cannot fire.**
Default `min_step_size = 1e-100` with `step_size_fail_factor = 0.5`: monotone halving from
0.1 would cross the floor in ~332 failures. After **66857** failures the stepsize was still
`1.86e-12`. So it is not descending — it is being restored between failures. Measured at the
same instant: `|t| = 5.82e-12`, `stepsize = 1.86e-12`, i.e. `stepsize ~ |t|/3.1`, which looks
like the stepsize is recomputed relative to time-to-go each iteration, discarding the
shrinkage. (That last step is an inference from the numbers; I have not located the reset in
the source.)

**3. Precision never escalates, so `MaxPrecisionReached` cannot fire either** -- and more to
the point, the thing that might let the step SUCCEED never happens. Measured:
`PrecisionChanged` fired exactly **once** in 3.9M failures, with `current_precision = 20`
and `condition number = 3.24e+08`.

[...]

PR #411, opened 2026-09-11
fix(trackers): a terminal step failure ends the path instead of looping forever

ofloveandhate:
Fixes the non-terminating tracking loop reported in #410.

## The bug

`MinimizeTrackingCost` throws when no `(precision, stepsize)` pair anywhere in
`[min_precision, max_precision]` satisfies criterion B.  Both AMP adjusters catch
that throw and report `SuccessCode::FailedToSelectPrecisionAndStepsize`:

* `AMPCriterionError()` — on a failed step
* `AdjustAMPStepSuccess()` — on a successful one

The throw happens **before either output is assigned**, so `current_precision_`
and `current_stepsize_` survive the failed step exactly as they were.

The stepping loop had no notion of a fatal step code — it saw a non-`Success`
return, called `OnStepFail()`, and went around again.  With nothing adjusted the
retry is **bit-identical**: `PreIterationCheck` sees the same precision and the
same stepsize, passes, and the identical step runs again.  Forever.

Observed on a real run (a cellular decomposition of the Whitney umbrella, one
fiber track during interslice): a single `Solve()` that never returned, **750k
failed steps in 135 seconds** at a frozen precision of 20 and a frozen stepsize
of `1.862645e-12`, successes frozen at 343, one core pegged.  Every printed digit
identical across the whole sampling window.  Killed at 90 minutes; three separate
runs behaved identically.

The contract was already documented — in `AdjustAMPStepSuccess`:

> The throw inside MinimizeTrackingCost remains semantically correct: it fires
> only when no valid pair exists anywhere in [p_min, p_max], indicating that the
> path has entered a region that cannot be tracked at any supported precision.
> **That is a genuine path failure, and is reported via
> `SuccessCode::FailedToSelectPrecisionAndStepsize`.**

It is reported.  The loop simply never listened.

## The fix

A `StepFailureIsTerminal()` predicate, and a loop that honors it: a terminal code
ends the path with that code instead of retrying.  Terminal means *the tracker
took no corrective action*, so the retry cannot differ from the attempt that just
failed.  Currently the sole member is `FailedToSelectPrecisionAndStepsize`; the
predicate exists so the set can grow with one line and a reason.

`base_tracker.hpp` only — no config change, no behavior change on any path that
was making progress.

## Tests

* `amp_tracker_terminates_when_no_precision_stepsize_pair_exists` — empties the
  scan window on purpose (a tolerance demanding ~40 digits against
  `maximum_precision = 20`) and requires `TrackPath` to **return** the failure.
  It carries `* boost::unit_test::timeout(60)` so a regression reports as a test
  failure rather than wedging CI, which is what this bug does to a build.

  That it exercises the new path is provable rather than assumed:
  `FailedToSelectPrecisionAndStepsize` is produced only inside
  `TrackerIteration` (`amp_tracker.hpp:912` and `:1037`), never by
  `TrackerLoopInitialization` or `PreIterationCheck`, so the only way `TrackPath`
  can *return* it is through the new early return.

* `terminal_step_codes_are_exactly_the_ones_that_adjust_nothing` — pins the
  predicate's membership, so widening it later is a deliberate edit.

All 104 cases in `test_tracking_basics` pass.

## Not in this PR

The other half of #410 stands on its own and is untouched here: **`max_num_steps`
counts only successful steps**, so a failing path is un-budgeted.  With this fix
the observed path fails cleanly instead of spinning, but a path that fails
millions of steps for some *other* reason still has nothing to stop it.  That
wants a budget decision (and probably a config field), so it belongs in its own
change.

### DIFF
--- a/core/include/bertini2/trackers/base_tracker.hpp
+++ b/core/include/bertini2/trackers/base_tracker.hpp
@@ -50,6 +50,32 @@ namespace bertini{
 	namespace tracking{
 
 		/**
+		\brief Whether a failed step leaves the tracker with no corrective action taken.
+
+		The stepping loop answers an ordinary failed step by adjusting precision and/or
+		stepsize and trying again, so the retry differs from the attempt that failed.  A
+		few outcomes admit no such adjustment: the tracker's state is left exactly as it
+		was, and retrying would repeat the identical step forever.  Those outcomes are
+		terminal -- the path must fail with that code instead of being retried.
+
+		Currently the sole member is \c FailedToSelectPrecisionAndStepsize, which the AMP
+		adjusters report when MinimizeTrackingCost finds no valid (precision, stepsize)
+		pair anywhere in the allowed window.  It assigns neither, so \c current_precision_
+		and \c current_stepsize_ survive the failure unchanged.
+
+		\param c The code returned by a tracker iteration.
+		\return True if the tracking loop must stop rather than retry.
+
+		\see Tracker::TrackPath
+		*/
+		inline
+		bool StepFailureIsTerminal(SuccessCode c)
+		{
+			return c == SuccessCode::FailedToSelectPrecisionAndStepsize;
+		}
+
+
+		/**
 		\class Tracker
 
 		\brief Base tracker class for trackers offered in Bertini2.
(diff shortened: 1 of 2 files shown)


ofloveandhate:
good to have this infinite loop closed!

PR #413, opened 2026-09-13
fix(system): MakeMovingHomotopy decides row identity on the canonical encoding, not a 6-digit rendering

### LINKED ISSUE #391: MakeMovingHomotopy: 'identical moving row' guard compares 6-digit renderings, so distinct rows are falsely refused
## Summary

`MakeMovingHomotopy` decides whether two moving rows are "identical" by comparing the
strings that `operator<<` produces for them. That stream uses the default `ostream`
precision — **6 significant digits** — so two genuinely different linear forms that agree
to 6 digits are declared identical and the homotopy is **refused**:

```
MakeMovingHomotopy: moving row 0 (`0.735966*x+(-2408.97)+0.593277*y+(-0.890232)*z`)
is identical in start_moving and end_moving, so it does not move; put non-moving
equations in `fixed` instead.
```

The threshold is **relative**: ~1e-6 of the constant's magnitude. At coefficients of order
1 it is 1e-6 and invisible; at order 2400 it is ~2.4e-3, which is easily inside the range
of legitimately distinct slice values. So this only bites at larger coordinate scales,
which is what makes it look sporadic.

This is a **false positive in a guard that aborts a valid computation**.

## Reproducer

```python
import bertini, bertini.nag_algorithm as na

bertini.default_precision(30)
x, y, z = bertini.variables(list('xyz'))

fixed = bertini.System(); fixed.add_variable_group([x, y, z])
fixed.add_functions([x**2 - y**2*z])

def lin(const):
    S = bertini.System(); S.add_variable_group([x, y, z])
    S.add_functions([bertini.coefficient('0.735966')*x
                     + bertini.coefficient('0.593277')*y
                     - bertini.coefficient('0.890232')*z
                     - bertini.coefficient(const)])
    return S

for d in ('1e-12', '1e-8', '1e-5', '1e-3', '1e-2', '1e-1'):
    v2 = repr(2408.97 + float(d))
    try:
        na.moving_homotopy(fixed, lin('2408.97'), lin(v2))
        print(f"delta={d:>6}  moved OK")
    except RuntimeError as e:
        print(f"delta={d:>6}  REFUSED: {'identical' in str(e)}")
```

Output:

```
delta= 1e-12  REFUSED: True
delta=  1e-8  REFUSED: True
delta=  1e-5  REFUSED: True
delta=  1e-3  REFUSED: True      <-- genuinely different rows, 1e-3 apart
delta=  1e-2  moved OK
delta=  1e-1  moved OK
```

A delta of `1e-3` at a constant of `2408.97` is a perfectly meaningful separation — those
are two different slices — and it is refused.

## Cause

`core/src/system/system.cpp` (~1840):

```cpp
auto function_strings = [](System const& s) {
    std::vector<std::string> out;
    for (auto const& f : s.NaturalFunctionsAsNodes())
    {
        std::ostringstream ss;
        ss << f;                       // default precision: 6 significant digits
        out.push_back(ss.str());
    }
    return out;
};
...
for (size_t i = 0; i < start_funcs.size(); ++i)
    if (start_funcs[i] == end_funcs[i])
        throw std::runtime_error("MakeMovingHomotopy: moving row ... is identical ...");
```

The same `function_strings` output also backs the check just above it (the "function appears
in both the fixed system and the moving rows" throw), so **both guards share the defect**: a
fixed function that merely *renders* like a moving one would also throw spuriously.

## Why this is the wrong tool, by the project's own rule

`docs/adr/0042` and the repo's own guidance are explicit that the canonical encodings are
digest **preimages, never presentation**, and that human-readable renders are derived views
free to change. `operator<<` is presentation. Using it to make an **identity decision** is
exactly the inversion that rule forbids — and it silently couples a correctness guard to a
formatting default.

The exact-value encoder this needs already exists in tree:

```cpp
// core/include/bertini2/function_tree/canonical_encoding.hpp
std::string CanonicalEncoding(std::shared_ptr<Node> const& root);
```

## Suggested fix

1. Compare with `CanonicalEncoding(f)` instead of `ss << f` in `function_strings` — exact
   values, no formatting dependence. Keep `operator<<` for the *message* text, which is
   presentation and appropriate there.
2. Apply to both guards (the fixed-vs-moving check and the start-vs-end check).
3. Consider whether string identity is the right predicate at all: two structurally different
   but mathematically identical rows still pass. Evaluating `start - end` at a random point
   would be a stronger test. At minimum, exact encoding removes the false positives.
4. A regression test with two rows differing at the 1e-3 level on a ~1e3-magnitude constant
   would have caught this; the scale dependence is the part that makes it easy to miss.

## Impact — measured on the real failure, not just the synthetic case

Hit in a surface cell decomposition (whitney `x^2 - y^2 z`, a component whose bounding
sphere has radius ~1275, so projection values run to a few thousand). A slice curve's fiber
move between two adjacent critical values was refused, aborting the whole decomposition.

I caught the refusal and evaluated `start_moving` and `end_moving` at two independent random
points to check whether the rows really were identical:

```
trial0 row0:  start = -2404.4890203631344
              end   = -2404.4890010320519
              DIFFERENCE = -1.933108e-05   (relative 8.04e-09)

trial1 row0:  start = -2406.9979659498613
              end   = -2406.9979466187788
              DIFFERENCE = -1.933108e-05   (relative 8.03e-09)

rendered start: 0.735966*x+(-2408.97)+0.593277*y+(-0.890232)*z
rendered end  : 0.735966*x+(-2408.97)+0.593277*y+(-0.890232)*z
```

The two rows differ by a consistent `1.93e-05` and render identically. The caller was moving
between two distinct slices; the guard asserted otherwise and killed the run.

## Environment

- b2 `VERSION` 3.5.0.dev0, branch `feature/eval-precision-tolerant` @ `c24bdc2e`
- installed dist `bertini2` 2.0.2
- Python 3.14.5, Linux aarch64

ofloveandhate:
Fixes #391.

Both of `MakeMovingHomotopy`'s identity guards compared functions via `operator<<`.
That is a **presentation** rendering — the default ostream precision is 6 significant
digits — so two genuinely different moving rows that happened to agree to 6 digits
compared *equal*, and a valid deformation was refused with a message asserting its two
endpoints were the same row.

The collision is ~1e-6 **relative**, so it bites at every coordinate scale rather than
only large ones. Measured refusals at rows 1e-3 apart on a constant of 2409, and
1.6e-7 apart on a constant of 0.163.

Found by a surface cell decomposition whose two adjacent slice values were 1.93e-05
apart at magnitude 2409 — distinct by ~7 orders of magnitude relative to their own
1.8e-12 accuracy. The caller was right; the guard was wrong.

## Fix

Identity now uses `node::CanonicalEncoding`, the exact-value encoding the content
digests are built on. This is the preimage-not-presentation distinction that ADR-0042
exists to enforce, applied to an identity decision that had quietly been coupled to a
formatting default. `operator<<` is retained for the human-readable message text.

Fixed in **both** guards, not only the one that bit — the
fixed-function-duplicated-in-moving-rows check shared the same comparison.

## Tests

Regression test in `core/test/classes/moving_homotopy_test.cpp` covering rows that
agree to 6 digits but differ exactly.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,24 @@ _______________________________________________________________________________
 
 _______________________________________________________________________________
 
+## [3.5.0] - unreleased
+
+A correctness fix to the `MakeMovingHomotopy` guards: they decided function identity on a
+*presentation* rendering, which silently refused valid homotopies.
+
+### Fixed
+
+- `MakeMovingHomotopy` no longer rejects a valid deformation whose two moving endpoints merely
+  *print* the same.  Both of its guards (a fixed function duplicated in the moving rows, and a
+  moving row identical at both endpoints) compared functions via `operator<<`, whose default
+  stream precision is **6 significant digits** -- so two genuinely different rows agreeing to 6
+  digits compared equal and the homotopy was refused with a message asserting the endpoints were
+  the same row.  The collision is ~1e-6 *relative*, so it bit at every coordinate scale.  Identity
+  is now decided on `node::CanonicalEncoding` -- the exact-value encoding the content digests are
+  built on (ADR-0042) -- while `operator<<` is still used for the human-readable message text.
+  Found by a surface cell decomposition whose slice values were 1.9e-5 apart at a magnitude of
+  2409, and reproduced at order one (0.162749 vs 0.1627491).  (bertiniteam/b2#391)
+
 ## [3.4.0] - 2026-07-16
 
 A one-call way to build a linear slice that passes through a chosen point, an endgame-hardening
--- a/core/src/system/system.cpp
+++ b/core/src/system/system.cpp
@@ -29,6 +29,7 @@
 #include "bertini2/system/system.hpp"
 #include "bertini2/system/slice.hpp"   // for System::Slices() (needs the full Slice type)
 #include "bertini2/function_tree/find.hpp"
+#include "bertini2/function_tree/canonical_encoding.hpp"  // exact-value identity (issue #391)
 
 #include <algorithm>
 #include <sstream>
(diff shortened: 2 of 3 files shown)


ofloveandhate (commented):
yep.  spot-on.  I really do like the album i'm listening to.  Musca, dang, it's sooo good.

Issue #27, opened 2016-05-11
Bertini hangs when run in parallel


adamjstewart:
I'm trying to install Bertini for a user and I'm running into a problem. After downloading the MPI2 version of Bertini for 64 bit Linux (either 1.4.1 or 1.5) and extracting it, I try the following commands:
```bash
mpirun -np 1 bertini -help
mpirun -np 2 bertini -help
```
The first command completes successfully. However, when I run the second command, it prints the help message, but afterwards hangs and never returns control to the command line. Is this expected behavior? I'm not sure if there's a better way to test the software than this.

aliddell:
This is the expected behavior. From the original developer: "only the head node check for the '-help' command and exits while the workers wait to perform computations." This is to show that the individual workers are created (i.e., MPI is working properly).

adamjstewart:
@aliddell: Thanks for the good news!

PR #414, opened 2026-09-13
feat(bindings): runtime get/set for PowerSeriesConfig and CauchyConfig on endgame objects

ofloveandhate:
The flavor-specific endgame configs were **constructor-only**. Once an endgame object
existed — for example the one a `HomotopySolver` builds internally — there was no way to
reach `PowerSeriesConfig.max_cycle_number`, `CauchyConfig.maximum_cauchy_ratio`,
`num_consecutive_same_cycle_number` and friends. That is unlike the generic
`EndgameConfig` / `SecurityConfig`, which already have `get/set_endgame_settings` and
`get/set_security_settings` on `EndgameBaseVisitor`.

## Change

* `get/set_powerseries_settings` on `PowerSeriesVisitor`
* `get/set_cauchy_settings` on `CauchyVisitor`

mirroring the existing generic accessors (`Configured::Get<T>` / `Set<T>`). `get` returns
a detached copy (`copy_const_reference`); mutate it and hand it back through `set`. The
pure-Python endgame wrapper forwards these via `__getattr__`, so they are reachable on the
`bertini.endgame.*` classes directly.

## Why

This unlocks tuning the singular-endpoint gates from Python at runtime — in particular
running a tracked connection through Cauchy with a permissive `maximum_cauchy_ratio`
instead of a track-then-guess fallback. Part of #27.

## Tests

Round-trip, detached-copy, and mutated-endgame-still-runs, in both flavor test files.
Full endgame suite (27 tests) green.

---

### DIFF
--- a/python/test/tracking/cauchy_endgame_test.py
+++ b/python/test/tracking/cauchy_endgame_test.py
@@ -270,3 +270,47 @@ def test_amp_cauchy_cycle_num_2(quadratic_homotopy, precision):
     assert code == SuccessCode.Success
     assert mp.abs(fa[0] - mpfr_complex(1)) < 1e-5
     assert eg.cycle_number() == 2
+
+
+# ---------------------------------------------------------------------------
+# flavor-config runtime accessors (get/set_cauchy_settings)
+# ---------------------------------------------------------------------------
+
+def test_amp_cauchy_flavor_config_setters(quadratic_homotopy):
+    """get/set_cauchy_settings round-trips, returns a DETACHED copy, and a mutated
+    endgame still runs.  Regression for the constructor-only flavor-config gap:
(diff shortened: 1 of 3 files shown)


python/test/tracking/cauchy_endgame_test.py
ofloveandhate, line 1: I still find the manipulation of tracker and algorithm settings awkward, as this test code proves.  Future work.  What's written here is decent, given the interface.  I wonder if the got settings, should be a reference and not a copy, so that modifications make their way in?  still, this is ok.

ofloveandhate (commented):
Not my absolute fave interface to have arrived at -- it's verbose.  But i'll take it, and continue to work on the interface in the future.

PR #415, opened 2026-09-13
test(zero_dim): cross-software cyclic-5 regression against a Bertini 1 oracle; ADR-0029/0030

ofloveandhate:
Two independent pieces that were sitting on the same branch.

## 1. Cross-software cyclic-5 regression against a Bertini 1 oracle

Commits the 70 finite solutions of cyclic-5 as computed by **Bertini 1.7.0** — an
independent solver, at slightly-tighter-than-default tolerances — and asserts that
bertini2's adaptive solve finds exactly that set: a bijection within 1e-6.

Compared as a **point set**, deliberately. The finite solutions are variety points and
therefore invariant under the random homotopy, so path-by-path identity across solvers is
meaningless; only the set is a shared fact.

This catches a silently lost root (the count comes back short) or a wrong endpoint — the
failure mode ADR-0017 and the cyclic-5 diagnostic flagged. Marked `timeout(600)` like the
`solve_cyclic` example, since it runs the same heavy adaptive solve that overran the global
180s budget on Windows CI.

## 2. Two ADRs, plus an index backfill

* **ADR-0029** — the gamma-trick constant is a unit-modulus complex generated at
  `MaxPrecisionAllowed()`. A `node::Float` caps at its creation precision, and |gamma| = 1
  is about conditioning rather than genericity.
* **ADR-0030** — fixed-multiple solves use one uniform, config-sourced ambient precision,
  and the `ZeroDim` factory defaults to adaptive.

The `docs/adr/README.md` index was stale at 0021; this adds 0022–0028 as well and
cross-references ADR-0017.

Happy to split these into two PRs if preferred — they share only a branch, not a subject.

---


ofloveandhate:
Closing: both pieces here (ADR-0029/0030 and the cyclic-5 Bertini 1 comparison test) are already on develop, landed via followup/adrs_and_cyclic5_oracle. This branch forked 512 commits ago, so merging it as it stands would also delete 29 newer ADR files. Nothing is lost by closing.

PR #416, opened 2026-09-13
fix(trackers): the step budget counts failed steps, not only successful ones

ofloveandhate:
Closes the second half of #410. (The first half — a terminal step code the stepping loop
ignored — is #411.)

## The bug

`SteppingConfig::max_num_steps` is documented as *"the maximum number of steps allowed
during tracking"*, but both `PreIterationCheck` implementations compared it against
`num_successful_steps_taken_` alone:

```cpp
if (num_successful_steps_taken_ >= Get<Stepping>().max_num_steps)
    return SuccessCode::MaxNumStepsTaken;
```

So a path whose steps **fail** never approached its budget. It could fail indefinitely
while sitting at 0% of its allowance, because the only counter the guard watched was the
one that had stopped moving.

Observed on a real run: **3.9 million failed steps against 343 successful ones**, thirty
minutes at 100% CPU with no progress, and the enclosing `Solve()` never returning. Every
other exit was structurally unable to fire — the stepsize was not descending toward
`min_step_size`, precision was not escalating toward `maximum_precision` — so nothing could
stop that path at all.

## The fix

Use `NumTotalStepsTaken()`, which already existed and already returns
`num_failed_steps_taken_ + num_successful_steps_taken_`. A failing path now terminates with
`MaxNumStepsTaken`, which is what a path that has spent its whole allowance without
arriving should report.

Fixed in **both** trackers, since both carried the same comparison. The config comment now
states the contract where the setting is configured, rather than leaving "steps" to be read
as "successful steps".

**No behaviour change for any path making progress.** A path that reaches its endpoint in
far fewer than `max_num_steps` total is unaffected, and the default allowance is 1e5. All
105 cases in `test_tracking_basics` pass.

## Test

A path where every step fails on purpose — tracking tolerance 1e-30 is unreachable in
double precision, so the corrector never converges — against `max_num_steps = 10`.

The case **discriminates**, which is the point: counting only successes it never reaches
the budget and instead halves its stepsize ~330 times down to the `1e-100` `min_step_size`
floor, returning `MinStepSizeReached`; counting every step it returns `MaxNumStepsTaken`
after 10. So the asserted code distinguishes the fix from its absence, with ~320 steps of
margin.

## An honest negative result

This fix did **not** recover the case that motivated looking for it. Five decompositions in
a 40-seed block were timing out at 2400s, and I expected them to be un-budgeted failing
paths. With the fix live and verified through the Python bindings (a failing path now
returns `MaxNumStepsTaken`), **all five still time out**. So their cost is spread across
many paths rather than concentrated in one runaway path, and that is a separate problem.

The fix still closes a real hole — the 3.9-million-failed-step path above was real, and
nothing in the library could have stopped it — but it should not be advertised as a
performance fix.

## Noted, deliberately not changed here

The member `num_total_steps_taken_` is **vestigial** — declared and reset, never
incremented anywhere in the tree — while `NumTotalStepsTaken()` computes the sum instead.
It reads like usable state and is always zero. Worth removing, but it isn't this fix.

---

### DIFF
--- a/core/include/bertini2/trackers/amp_tracker.hpp
+++ b/core/include/bertini2/trackers/amp_tracker.hpp
@@ -616,7 +616,10 @@ namespace bertini{
 			*/
 			SuccessCode PreIterationCheck() const override
 			{
-				if (num_successful_steps_taken_ >= Get<Stepping>().max_num_steps)
+				// The budget counts EVERY step, not only the successful ones -- see the
+				// same check in FixedPrecisionTracker and issue #410.  This is the guard
+				// that bounds a path whose steps fail without ever advancing.
+				if (NumTotalStepsTaken() >= Get<Stepping>().max_num_steps)
 					return SuccessCode::MaxNumStepsTaken;
 				if (current_stepsize_ < Get<Stepping>().min_step_size)
 					return SuccessCode::MinStepSizeReached;
--- a/core/include/bertini2/trackers/config.hpp
+++ b/core/include/bertini2/trackers/config.hpp
@@ -111,7 +111,7 @@ namespace tracking{
 		unsigned consecutive_successful_steps_before_stepsize_increase = 5; ///< What it says.  If you can come up with a better name, please suggest it.  StepsForIncrease
 
 		unsigned min_num_steps = 1; ///< The minimum number of steps allowed during tracking.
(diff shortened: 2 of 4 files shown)


ofloveandhate:
another infinite loop bug, closed.  huzzah

Issue #417, opened 2026-09-13
n-ary canonicalization re-renders every operand to a string on each construction, making incremental sum/product building quadratic


ofloveandhate:
## Summary

Building an n-ary node (`SumOperator` / `MultOperator`) canonicalizes its operands, and the
canonicalization **recomputes a string render of every operand, every time**. Since nodes are
immutable and interned, that work is repeated rather than cached — so accumulating a sum one
term at a time is **quadratic in the number of terms, times the operand subtree size**.

Measured: building a 3200-term sum by repeated `+` takes **7.98 s**; building the same sum by
balanced pairwise reduction takes **0.069 s**. Same result, **116x**.

## Where

`core/src/function_tree/canonical.cpp`, `CanonicalizeNaryOperands` (~line 224). For an
N-operand node, per construction:

```cpp
for (auto const& op : operands)
    for (auto const& v : GatherVariables(op))     // walks every operand's subtree
        ...
std::sort(vars.begin(), vars.end(), ...);

for (std::size_t i = 0; i < N; ++i)
{
    keys[i]   = operands[i]->MultiDegree(vars);
    prints[i] = PrintOf(operands[i]);             // renders every operand to a string
}
```

`PrintOf` is used as a sort key for the canonical ordering. It is recomputed for every
operand on every construction, and it walks the operand's whole subtree.

So `total = total + term`, repeated N times, costs

    sum over i of  O( i * subtree_size )   =   O(N^2 * subtree)

and the subtree factor is what makes it bite: the same loop is cheap on small operands and
brutal on large ones.

## Reproducer (pure Python, via the bindings)

```python
import bertini, time
x, y = bertini.variables(['x','y'])
terms = [ (x*y)**(i % 7 + 1) * (x + bertini.coefficient(i)) for i in range(3200) ]

t = time.time()
tot = None
for u in terms: tot = u if tot is None else tot + u      # linear accumulation
print('linear   %.3f s' % (time.time()-t))

def balanced(ts):
    cur = list(ts)
    while len(cur) > 1:
        nxt = [cur[i] + cur[i+1] for i in range(0, len(cur)-1, 2)]
        if len(cur) % 2: nxt.append(cur[-1])
        cur = nxt
    return cur[0]

t = time.time(); balanced(terms); print('balanced %.3f s' % (time.time()-t))
```

    N       linear      balanced     speedup
    100      0.009 s      0.001 s        7x
    200      0.031 s      0.003 s       11x
[...]

PR #418, opened 2026-09-14
perf(function_tree): MultiDegree memoizes by node identity within one traversal

ofloveandhate:
Part of #417. The traversal-level memo only — measured, and the safe half.

## The problem

The node graph is hash-consed, so a small DAG can stand for an enormous expansion. But
`MultiDegree` is a plain recursive walk, so it visits the **tree**: O(expansion) where the
structure is O(DAG). `CanonicalizeNaryOperands` calls it for every operand of every n-ary
construction. Measured on a nested expression whose DAG grows ~3 nodes per level while its
expansion doubles:

    level   DAG nodes   printed length   MultiDegree   ONE multiply
        7          25            1,527       0.05 ms        0.09 ms
       14          46          196,599       8.22 ms       13.05 ms

A 46-node expression cost 13 ms to multiply by a variable. In a cofactor expansion inside an
isosingular deflation stage, that compounded into a stage that ran **4.5 hours** with a DAG
that never exceeded 1102 nodes.

## The change

`MultiDegree` becomes a non-virtual wrapper owning a **per-traversal** memo keyed on node
identity; the per-type work moves to `MultiDegreeImpl`, which recurses through the wrapper
so children are memoized. A shared subtree is visited once however many times it is reached.

A multidegree is only meaningful relative to a variable group, so a nested call asking about
a *different* group bypasses the memo rather than reading a wrong answer. Within one
traversal the same reference is handed down, so identity is the right test. The memo is
thread-local and released by RAII guard however the outermost call exits.

## Measured

18-level benchmark, same machine, same binary shape, only the change flipped:

    unmemoized    510 ms
    memoized      293 ms      1.74x

All `test_classes` pass **including the golden digest fixtures** — canonical ordering is
unchanged, which is the contract that matters (canonicalization feeds `CanonicalEncoding`).
The regression test pins a 420 ms budget between the two measured times.

## What was tried and rejected, so nobody repeats it

* **Caching the result across calls** (per-node `CachedMultiDegree`): **29% slower**
  (659 ms). Filling such a cache costs the same walk it avoids, so it only pays from a
  node's second mention. The memo has to be inside the traversal.
* **`Find` / `GatherVariables`** need nothing — already DAG-aware (they carry a `visited`
  set).

## Not in this PR

`PrintOf` — the canonicalization tie-break — is now the last traversal that walks the tree.
Computing it on demand (only on a multidegree tie) measured a further 293 → 199 ms, but I
reverted it after an endgame test failed with it in place. That test has since proven
**flaky** (fails, then passes 3/3 on the same binary), so the lazy print is very likely
fine; it belongs in a follow-up with its own measurement rather than bundled here. Making
`PrintOf` itself DAG-aware is the remaining fix for #417 proper.

---


ofloveandhate (commented):
yep.  memoization to fix a string printing problem.  the pattern is correctly implemented.

PR #422, opened 2026-09-15
fix(function_tree): a variable-free power is a constant -- homogeneous whatever its exponent (#419)

### LINKED ISSUE #419: Homogenize() silently skips a system with a variable-free sqrt subtree (an algebraic constant); AutoPatch then throws
## Summary

A system whose functions contain a **variable-free `sqrt` (or other operator) subtree** — a
constant computed symbolically, such as the golden ratio `phi = (sqrt(5) + 1) / 2` — builds,
evaluates and differentiates correctly, but `System::Homogenize()` does not homogenize it,
and the next `AutoPatch()` throws:

```
RuntimeError: requesting to AutoPatch a system which is not homogenized.  Homogenize it first.
```

Nothing tells the caller that `Homogenize()` did nothing.

## Reproducer (Python bindings)

```python
import bertini
x, y, z = bertini.variables(list('xyz'))
phi = (bertini.sqrt(bertini.coefficient(5)) + bertini.coefficient(1)) / bertini.coefficient(2)
f = (bertini.coefficient(4) * (phi**2 * x**2 - y**2) * (phi**2 * y**2 - z**2) * (phi**2 * z**2 - x**2)
     - (bertini.coefficient(1) + bertini.coefficient(2) * phi) * (x**2 + y**2 + z**2 - 1)**2)   # Barth sextic
s = bertini.System(); s.add_variable_group([x, y, z]); s.add_functions([f])
print(list(s.degrees()))          # [6]  -- the degree is read correctly
s.homogenize()
s.auto_patch()                    # RuntimeError: ... not homogenized
```

Replacing `phi` by a 300-digit numeric coefficient makes everything work, so the system
is polynomial in every sense that matters; only the homogenizer's notion of "polynomial"
disagrees.

## Where it bit

bertini_real's `test/surface/barth6/input` writes `phi=(5^(1/2)+1)/2;` as a constant.
Porting it to bertini2 with `phi` as a node, the cellular decomposition ran its
critical-curve witness stage fine and died at the first projective move (b2_experiments,
`nid/slicing.py::projectivize_move`, 2026-09-15).  Every surface with an algebraic
irrational in its coefficients is affected.

## Expected

A subtree with no variables is a constant.  `Homogenize()` should either treat it as one
(degree 0) and homogenize, or refuse loudly with an error naming the offending subtree —
not return with the system unchanged.  `IsPolynomial()` (or whatever gate `Homogenize`
consults) should classify by *variable dependence*, not by operator type.

ofloveandhate:
Closes #419.

## What was wrong

`PowerOperator::IsHomogeneous` (both overloads, `core/src/function_tree/operators/arithmetic.cpp`)
answered "not homogeneous" for any power whose exponent is not a non-negative integer, without
asking whether the *base* involves the variables. `PowerOperator::Degree` already answers 0 for
the same node when the base is variable-free. So `5^(1/2)` -- Bertini 1's spelling of a square
root, and how `bertini_real`'s `barth6` input writes the golden ratio, `(5^(1/2)+1)/2` --
reported degree 0 and polynomial, but not homogeneous. `Homogenize()` pads by degree, found
nothing to do, and `AutoPatch()` then refused the result with "requesting to AutoPatch a system
which is not homogenized".

The issue's own reproducer used `bertini.sqrt(5)`, which builds a `SqrtOperator` and was already
classified correctly -- which is why that reproducer does not fail. The cellular decomposition
hit the power form.

## The fix

A power whose degree in the variables asked about is 0 is a constant and is homogeneous,
whatever its exponent. One guard per overload, placed before the existing integer-exponent
reasoning, so the two verdicts agree.

## Tests (C++, `homogenization_test.cpp`)

Four cases, table-driven, so a future operator that breaks the agreement fails by name:

- every operator's variable-free form -- `sqrt(5)`, `exp(1)`, `log(2)`, `sin(1)`, `cos(1)`,
  `tan(1)`, `5^(1/2)`, `5^(3/2)`, `(1+sqrt(5))/2`, `(5^(1/2)+1)/2`, `sqrt(2)^3`, `exp(sqrt(2))`,
  `-sqrt(3)`, `sqrt(pi)` -- is degree 0 (total, per variable, per group), polynomial,
  homogeneous, and as a coefficient leaves a polynomial's degree and homogeneity alone;
- a system with each of those as a coefficient homogenizes, gains its homogenizing variable, and
  auto-patches;
- the Barth sextic with the golden ratio as a node, in the `5^(1/2)` spelling: degree 6,
  homogenizes, patches;
- the controls with a variable inside (`sqrt(x)`, `exp(x)`, `log(x)`, `sin(x)`, `x^(1/2)`,
  `sqrt(x*y)`, `2^x`) stay non-polynomial and non-homogeneous.

Before the fix, the `5^(1/2)`, `5^(3/2)` and `(5^(1/2)+1)/2` rows fail (measured through the
Python bindings against the unfixed library: `is_homogeneous` False, `auto_patch` raises);
everything else already passed. After: the homogenization suite is 24 cases / 338 assertions
green, and the whole `test_classes` executable passes.

## Answering the question the issue raised

Does this happen for other node types? The sweep says no: `sqrt`, `exp`, `log`, the
trigonometric functions and their compositions were already consistent; only the rational power
disagreed with its own degree.


ofloveandhate (commented):
i was silly to have made this mistake with my work on the symbolic engine when i wrote it.  i am glad that it's more correct now.

PR #423, opened 2026-09-15
fix(function_tree): negation preserves its operand's degree; the non-polynomial degree rule lives once in UnaryOperator

ofloveandhate:
The twin of #397, in the other class that inherited the same mistake.

## What was wrong

`UnaryOperator::Degree(VariableGroup)` computed the group degree by **summing** the per-variable
degrees. That equals the total degree only for a single monomial. `PowerOperator` had inherited it
(#397, fixed in #398); `NegateOperator` still did, so

```
-(x^2 + y^2)   reported degree 4 in {x, y}   (true degree 2)
-(x^3 + x*y)   reported degree 4             (true degree 3)
```

while the total degree and the multidegree were right. `System::DegreeBound()`, which sizes
adaptive precision, followed the inflated number, and `SumOperator::Homogenized` pads by
`Degree(vars)`, so a negated sum could also be padded wrongly.

## The fix

- `UnaryOperator` now owns the non-polynomial rule once, in both overloads: an operand free of
  the variables asked about makes a constant (degree 0); anything else is not a polynomial
  (degree -1). This is exactly what `sqrt`, `exp`, `log` and the trigonometric operators each
  implemented in their own identical `Degree` override -- those four copies are deleted.
- `NegateOperator`, a degree-preserving unary like `IntegerPowerOperator`, overrides degree,
  group degree and multidegree to pass its operand's through.

Net: 113 insertions, 88 deletions, and the rule has one home.

## Tests (C++, `degrees_test.cpp`)

- `negation_of_a_sum_has_the_sum_degree`: the two cases above, the monomial control, one-variable
  and total degrees, the multidegree `[3, 1, 0]`, homogeneity of a function containing a negated
  sum, and `System::DegreeBound() == 2` (was 4).
- `non_polynomial_unary_operators_share_one_degree_rule`: `sqrt`, `exp`, `log`, `sin`, `cos`,
  `tan` of a variable-dependent operand are degree -1 and non-polynomial; of an operand that does
  not involve the group asked about they are constants; of a genuine constant they are constants
  in every sense.

Measured before the fix through the Python bindings: `(-(x**2+y**2)).degree([x,y]) == 4`.


ofloveandhate (commented):
another naive math mistake i made when writing the degree code.  glad it's fixed.  i looked at every file and they're correct.

PR #424, opened 2026-09-15
fix(trackers): the path-truncation check measures the largest coordinate, like the endgame and the finiteness threshold (#404)

### LINKED ISSUE #404: Tracker and endgame path truncation disagree: independent defaults (1e5 vs configured) and different norms (2-norm vs infinity)
## Summary

There are two independent path-truncation thresholds — one on the tracker, one on the
endgame — and they disagree in **default value**, in **norm**, and in **configuration**.
Raising one does nothing to the other, so a caller who deliberately loosens the endgame's
ceiling still has paths killed earlier by the tracker's default.

```cpp
// tracker  -> SuccessCode::GoingToInfinity
// core/include/bertini2/trackers/base_tracker.hpp:503
if (GetSystem().DehomogenizePoint(std::get<Vec<ComplexT>>(current_space_)).norm()
        > path_truncation_threshold_)

// core/include/bertini2/trackers/base_tracker.hpp:621
NumErrorT path_truncation_threshold_ = 1e5;

// endgame  -> SuccessCode::SecurityMaxNormReached
// core/include/bertini2/endgames/powerseries.hpp:809
norm_of_dehom_of_latest_approx = this->GetSystem().InfinityNormOfDehomogenized(latest_approx);
if (this->BeyondSecurityMaxNorm(norm_of_dehom_of_latest_approx) && ...)
```

## The two defects

**1. The settings are independent, so the endgame's ceiling can be silently overridden.**
`Security::max_norm` is exposed and documented as the divergence-bailout knob, and a
caller who raises it reasonably expects large-but-finite endpoints to survive.  But the
tracker truncates first, on its own `path_truncation_threshold`, which that caller never
touched.  Measured in a client that deliberately set `max_norm = 1e14` (its default of
1e4 was truncating genuine large-but-finite endpoints): the tracker sat at 1e5
throughout, so the decision never took effect and the paths died as `GoingToInfinity`
rather than surviving.  The client's workaround is to set the tracker's threshold to the
same value at every solve — which suggests the library should be relating them itself.

**2. The two use different norms.**  The tracker uses `.norm()` — the Euclidean 2-norm —
while the endgame (and the finiteness classifier) use the infinity norm.  So "the same"
threshold means different things in the two places, and the discrepancy grows with the
number of coordinates: for an n-coordinate point the 2-norm can exceed the infinity norm
by up to sqrt(n), so the tracker is systematically the stricter of the two even when the
numbers are set equal.

## Suggested shapes

- Have the tracker's truncation derive from, or share configuration with, the endgame's
  security settings, so there is one knob with one meaning.  If they must stay separate
  (they do serve different phases), then at minimum make the relationship explicit and
  documented, and consider defaulting the tracker's to the endgame's value.
- Use the same norm in both.  The infinity norm looks like the right choice, matching
  `endpoint_finite_threshold` and `Security::max_norm`.

## Related

bertiniteam/b2#403 — both of these thresholds are also taken over the WHOLE dehomogenized
point, synthetic coordinates included, which is that issue's subject.  Noted here only
because a fix to either one should probably be aware of the other; the two issues are
otherwise independent, and this one stands even if every coordinate is natural.

ofloveandhate:
Closes #404, as narrowed by its own follow-up comment: the separate defaults for the tracker and
the endgame look intentional, so what remained was the norm mismatch.

## What was wrong

Three thresholds are meant to answer one question -- "is this point too large to be finite?" --
and two of them measured the largest coordinate (the infinity norm) while the third measured the
2-norm:

| where | what it measured |
|---|---|
| tracker, `CheckGoingToInfinity` against `path_truncation_threshold` | **2-norm** |
| endgame, `Security::max_norm` | infinity norm |
| post-processing, `endpoint_finite_threshold` | infinity norm |

So a path whose coordinates all stayed under `path_truncation_threshold` was truncated by the
tracker once `sqrt(n)` carried its 2-norm over the line -- with the default `1e5`, a
three-coordinate point at `8e4` each is truncated (2-norm `1.39e5`) although no coordinate is
anywhere near the threshold, and the endgame would not have flagged it. Bertini 1 measures the
largest coordinate.

## The fix

`base_tracker.hpp`: the truncation check uses `lpNorm<Eigen::Infinity>()` on the dehomogenized
point. One expression.

Riding along, because it is the same file and was flagged in #416: the `num_total_steps_taken_`
member was declared and reset but never incremented (`NumTotalStepsTaken()` already returns the
sum of successes and failures). Deleted.

## Test (C++, `fixed_precision_tracker_test.cpp`)

`truncation_threshold_applies_to_the_largest_coordinate`: a straight path in three variables to
`(8e4, 8e4, 8e4)` with threshold `1e5` now tracks to `Success` and lands on the endpoint (it was
`GoingToInfinity` before the fix); a path whose *largest* coordinate reaches `1.2e5` is still
truncated.


ofloveandhate (commented):
yep.  glad they're all using the same norm now.  caused confusion!!!

PR #425, opened 2026-09-15
perf(linalg): random conjugate-orthonormal matrices come from a factorization sized to the shape requested (#401)

### LINKED ISSUE #401: RandomConjugateOrthonormalMatrix is O(max(rows,cols)^3) even when one dimension is tiny
`RandomConjugateOrthonormalMatrix` is O(max(rows,cols)^3) even when one dimension is tiny

`eigen_extensions.hpp:592` generates a random conjugate-orthonormal matrix by drawing a
SQUARE matrix of the larger dimension, QR-factoring it, and returning the leading
`rows x cols` block:

```cpp
const unsigned int dim = max(rows, cols);   // generate square, then truncate
Mat<NumberType> seed = RandomOfUnits<NumberType>(dim, dim);
Eigen::HouseholderQR<Mat<NumberType> > qr(seed);
Mat<NumberType> Q = qr.householderQ() * Mat<NumberType>::Identity(dim, dim);
return Q.topLeftCorner(rows, cols);
```

That is fine near-square, and catastrophic for an oblique shape.

**Measured**, squaring up an isosingular-deflated system (8 variables, 4908 functions
after one deflation stage appends C(8,4)^2 = 4900 minors):

`RandomConjugateOrthonormalMatrix<mpfr>(8, 4908)` builds a **4908 x 4908** mpfr matrix
(~24M entries) and Householder-QRs it -- on the order of **1.2e11** mpfr operations --
to keep the 8 x 4908 corner, i.e. 0.16% of the work. It does not finish inside 900
seconds at 30 digits. Stack (via SIGABRT faulthandler):

```
bertini/random/__init__.py:129 in random_matrix
  RandomConjugateOrthonormalMatrix<mpfr_float>
    Eigen::internal::householder_qr_inplace_blocked<Matrix<mpfr_float>>
```

When `rows <= cols` only ORTHONORMAL ROWS are required, which is a thin QR of a
`rows x cols` draw: O(rows^2 * cols) = 64 * 4908 ~ 3e5 operations here, ~400,000x
cheaper, and it yields exactly the same distribution of row-orthonormal matrices.
Symmetrically for `rows >= cols` (orthonormal columns, QR of the `rows x cols` draw).

The square-and-truncate recipe is documented as mirroring Bertini 1. Worth keeping the
guarantee (conjugate-orthonormal, well-conditioned, seed laundered by the QR) while
sizing the factorization to the shape actually requested.

Affects any caller that randomizes a very overdetermined system down to square --
isosingular deflation is the natural producer, since it appends C(m,k)*C(n,k) minors.

ofloveandhate:
Closes #401.

## What was wrong

`RandomConjugateOrthonormalMatrix(rows, cols)` followed Bertini 1's recipe literally: draw a
square random matrix of the *larger* dimension, QR-factor it, and keep the leading `rows x cols`
block. That is O(max(rows, cols)^3) whatever the other dimension is. An isosingular deflation
asked for an `8 x 4908` randomization matrix, which became a `4908 x 4908` multiprecision
factorization and did not finish in 900 s.

## The fix

Factor a matrix sized to the request: draw `max x min` (the longer side by the shorter), QR-factor
it to a Q with `min` orthonormal columns, and transpose when the requested shape is wide, so the
rows come out orthonormal (`M M^H = (Q^H Q)^T = I`). Cost O(max * min^2). The distribution is
unchanged: the leading block of a Haar-distributed unitary is a Haar-distributed orthonormal
frame, which is exactly what the thin factorization of a Gaussian-like draw produces.

A **square** request draws and factors exactly as before, so seeded square matrices are
bit-identical to what they were. Non-square requests consume fewer random draws than the old
recipe and therefore come out different for the same seed -- a different but equally valid
random matrix. Nothing in the test suites pins those draws.

## Tests (C++, `eigen_test.cpp`, suite `random_conjugate_orthonormal`)

- `shape_and_orthonormality_both_ways`: `3x7`, `7x3`, `5x5`, `1x6`, `6x1` at multiprecision have the
  requested shape and are orthonormal in rows (wide) or columns (tall/square) to `1e-25`; the
  double type too.
- `cost_scales_with_the_shape_requested`: `4 x 1000` against `4 x 2000` at multiprecision -- the
  ratio must stay under 4 (linear in the long side gives ~2; the old recipe was cubic, ~8, and its
  `2000 x 2000` factorization alone runs for minutes).


ofloveandhate (commented):
huzzah, i can make bigger matrices.  code does what it says, unit tests test what they are supposed to test.  this feels good to me.

PR #426, opened 2026-09-15
fix(io): the classic writer declares the path variable, so a homotopy round-trips (#366)

### LINKED ISSUE #366: to_classic_input() does not round-trip a system with a path variable (missing pathvariable declaration) — breaks move-homotopy replay
## Summary

`System::to_classic_input()` does not round-trip a system that carries a **path variable**.
It emits functions that *reference* the path variable but never emits the `pathvariable`
declaration, so `bertini.parse.system(sys.to_classic_input())` fails to parse the result.

Every system should round-trip through classic input; a homotopy (move) is a system too.
This breaks any records/replay layer that stores a solve by its classic-input text — a
recorded move-homotopy cannot be regenerated into a runnable system.

## Reproduction

```python
import bertini
from bertini import nag_algorithm as na
bertini.default_precision(30); bertini.random.set_random_seed(42)
x, y = bertini.variables(['x', 'y'])
s = bertini.System(); s.add_variable_group([x, y]); s.add_function(x**2 + y**2 - bertini.coefficient(1))
a = bertini.Slice.random_real([x, y], 1); b = bertini.Slice.random_real([x, y], 1)
homot = na.moving_homotopy(s, a.as_system(), b.as_system())   # has a path variable t
bertini.parse.system(homot.to_classic_input())     # -> FunctionParser parse error at the first 't'
```

The emitted INPUT section:

```
INPUT
variable_group x, y;
function f0, f1;
f0 = x^2+y^2-1;
f1 = ((-0.707107)*x+(-0.438321)+0.707107*y)*(1-t)+(0.963512,0.267666)*(...)*t;
END;
```

No `pathvariable t;` line, yet `t` appears in `f1`; `parse.system` sees an undeclared
symbol and (correctly) refuses it.

## Expected

`to_classic_input()` on a system with a path variable should emit the path-variable
declaration (`pathvariable t;`) in the INPUT section, so `parse.system(to_classic_input())`
reproduces an equivalent system — including a homotopy.

## Notes

- The `(re,im)` complex-coefficient literals in the same output are already handled by the
  recent parse.system tolerance work; the missing `pathvariable` declaration is the
  remaining blocker.
- Found via a cellular-decomposition records layer that stores every solve (including slice
  moves) by classic-input text for replay. Non-move solves round-trip fine; move-homotopies
  do not, so those records are stored-but-not-replayable. "Every solve record should be
  fully round-trippable" is the contract this breaks.
- Same family as #364 / #365: a capability present in the C++ objects but not faithfully
  exposed through the text round-trip the Python layer relies on.

ofloveandhate:
Closes #366.

## What was wrong

`EmitSystem` (`core/include/bertini2/io/classic_writer.hpp`) wrote the variable groups, the named
subexpressions and the functions, but never the path variable. For a homotopy the functions use
`t`, the file never declares it, and the text does not parse back. (Until #398 the failed parse
was silent -- an empty `System` came back; now it is an error. Either way the round trip was
broken.)

## The fix

One declaration, emitted with the variable groups when the system has a path variable:

```
pathvariable t;
```

That is the spelling the classic grammar reads (`system_rules.hpp`), so a homotopy now
round-trips through the writer and the parser like any other system.

## Test (C++, `classic_parsing_test.cpp`)

`classic_writer_round_trips_a_homotopy_with_its_path_variable`: a two-function homotopy in `x, y`
with path variable `t` emits `pathvariable t;`, parses back with its path variable and function
count intact, and evaluates identically at a generic point and time.

## What this does and does not touch

This is the emitter half of #366. The records archive never stored classic text (it stores the
canonical encoding), so nothing in `records/` changes; the slow path the issue measured lived in
a downstream tool's own checkpoints, which re-parsed classic text per solve and will move to the
canonical-encoding reader planned for 3.5.


ofloveandhate (commented):
yep

PR #427, opened 2026-09-15
fix(system): an affine products-of-linears block homogenizes, like the linear-forms block (#376)

### LINKED ISSUE #376: Products-of-linears start systems do not homogenize (ExpandToFunctionTree variable count mismatch)
## Symptom

Homogenizing a system whose functions include a product of linears (the two-step sphere-cap regeneration's `L1 * L2` start, and linear-product start structures generally) fails with

    ExpandToFunctionTree: products-of-linears variable count mismatch

## Why it matters

The cellular-decomposition reliability work established that complex-gamma moves must track **projectively** (homogenize + patch) so paths only go to infinity at the end, never mid-path (mid-path affine poles were diagnosed as `pre:GoingToInfinity` divergence). That fix was applied to the regeneration witness moves and collapsed mid-path divergence — but the **product-to-sphere deformation cannot be projectivized** because its products-of-linears start system will not homogenize, so that one move family still tracks affinely.

The measurable consequence: the residual preimage-curve cap losses in the whitney gauntlet (1–2 truncated paths per dirty run, `SecurityMaxNormReached`/`pre:GoingToInfinity` during `regen-slicemove`/`regen-linprod: sphere-cap`), each of which costs a critpointset a genuine point and produces mis-assembled faces downstream. Raising the security ceiling (max_norm/path_truncation 1e14) recovers most, but the principled fix is projective tracking for this move family, which is blocked here.

## Ask

Make homogenization handle products-of-linears function structures (or provide an expansion path so the homogenizer sees plain polynomials), so the sphere-cap deformation can join the projective-move policy.

ofloveandhate:
Closes #376.

## What was wrong

`ProductsOfLinearsBlock::Homogenize` was a no-op and `IsHomogeneous` always answered `true`. Both
were written for the m-homogeneous start system, which builds its block already homogeneous
(zero constant column, homogenizing variables among the columns). An **affine** block -- what
`add_products_of_linears` builds, and the shape of a regeneration deformation such as
`(x-1)(x+1)` -- met the no-op, was reported homogeneous, and the homogenized `System` then
carried a block whose column count no longer matched its variables. Expanding to nodes threw
"products-of-linears variable count mismatch", and the system could not be tracked
projectively; a downstream decomposition had to keep one family of moves affine and raise its
divergence ceilings to compensate.

## The fix

The block now does exactly what `LinearFormsBlock` does: `Homogenize` folds each factor's
constant column onto the homogenizing variable -- which the `System` prepends to the ordering, so
the constant column moves to the front -- bumps the variable count, and evaluates `M * [vars]`
with no augmenting 1 from then on. `IsHomogeneous` answers from the data: homogenized, or every
constant column exactly zero (the start system's form). "Homogenized" is derived from the shape
(`columns == num_vars`; a constructed block always has `num_vars + 1`), so there is **no new
state, no serialization change, and no change to the canonical encoding** -- `b2sysenc` is
untouched. A second affine group throws, as it does for the linear-forms block.

`System::NaturalFunctionsAsNodes` (the expansion) learned that a homogenized block has no
constant column; for affine blocks it builds exactly the trees it built before, so no existing
expansion changes.

## Tests (C++)

- `products_of_linears_block_test.cpp`: `(x-1)(x+1)` homogenizes to `(x-h)(x+h)` -- values and
  Jacobian at `(h, x, y) = (2, 3, 5)` in double and multiprecision, degree unchanged, variable
  count 3, a second call throws; a block built with zero constants reports homogeneous without
  being asked.
- `homogenization_test.cpp`: the regeneration shape -- the block beside a polynomial slice --
  homogenizes as a system, expands to nodes that evaluate identically to the block form at a
  generic projective point (this is the call that threw), and auto-patches.


ofloveandhate (commented):
precisely what needed to happen.  would i have written the exact same code?  probably not.  am i thrilled that this solves the problem?  absolutely.

Issue #409, opened 2026-09-11
is_singular is decided by a condition-number threshold, and misses solutions lying ON a positive-dimensional singular component


ofloveandhate:
## Summary

`is_singular` in `SolutionMetaData` is decided by

```cpp
// nag_algorithms/zero_dim_solve.hpp:2146
smd.is_singular = (smd.multiplicity > 1) || (smd.condition_number > cond_threshold);
```

Both terms miss the same case, and it is not an exotic one: **a solution lying ON a
positive-dimensional singular component of the system.** Such a point has multiplicity 1
(it is an isolated solution of the *sliced* system) and an unremarkable condition number
(the endgame converges cleanly to it), so it is reported **non-singular** — while the
Jacobian there is rank-deficient.

That is the case that matters when a caller uses `is_singular` to *separate* singular
structure from smooth structure, which is a natural and, as far as I can tell, intended use
of the flag.

## How it shows up

Downstream, in a numerical-irreducible-decomposition / cellular-decomposition application
built on b2: the critical curve of a surface is computed by regeneration, and its witness
set is split into singular and non-singular points, the singular ones seeding the singular
curves. The split was `if (md.is_singular)`.

On one random projection of the Whitney umbrella `x^2 - y^2*z`, whose singular locus is the
z-axis, a z-axis witness point was reported non-singular. The consequence cascaded:

```
z-axis classified non-singular
  -> the whole singular curve (17 vertices, 8 edges) duplicated into the critical curve
  -> critical-curve edge midpoints at (0, 0, z), where grad f = (2x, -2yz, -y^2) == 0
  -> a downstream homotopy's start point is rank 8 of 9
     (smallest singular value 9.3e-18 against 1.5e-01 at healthy points)
  -> 64 of 177 tracks return NeverStarted
  -> 91 validator problems, Euler characteristic 15 against a truth of 1
```

A *different* random projection of the same surface classified the same structure
correctly. Same code, same surface — so the flag is not merely conservative or
aggressive, it is unreliable in both directions.

## What works instead

Gate on the **numerical rank of the Jacobian** at the point, not on the condition number:
compare the rank against the expected `n - dim`. The decisive property is that the zero
being tested is **structural rather than approached** — on a singular component the
Jacobian vanishes *identically* in exact arithmetic, where a condition number merely grows
without bound as you approach a fold and therefore admits no correct threshold.

Measured separation on that seed, between the leaked points and the genuine critical-curve
points:

```
[...]

Issue #365, opened 2026-07-21
Midpath path-crossing results are not persisted in records, and flagged paths still report Success


ofloveandhate:
## Summary

Two related deficiencies around the endgame-boundary midpath (path-crossing) check:

1. **The midpath result is not persisted in the records store.** A solve whose paths the
   midpath check flagged is recorded identically to a clean one, so the flag cannot be
   recovered from a record — only from stdout at the moment it ran.
2. **A flagged path still reports `Success`.** Its `SolutionMetaData.endgame_success_code`
   is indistinguishable from a path that tracked cleanly, so no per-path consumer can
   tell that the library considers the solution suspect.

Together these make an unresolved path crossing invisible to every consumer that reads
records or per-path metadata — which is every consumer except a human watching a
terminal.

## Evidence

Bertini itself says the solutions may be wrong:

```
warning: 3 path crossing(s) detected at the endgame boundary remained unresolved after
2 re-track attempt(s); the affected solutions may be wrong.  Consider a higher-order
predictor or a tighter tracking tolerance.  See EndgameBoundaryMetadata().
```

The structured report exists and is good — `solver.endgame_boundary_metadata()` returns a
`MidpathCheckReport` with `passed`, `num_crossings_detected`, `num_resolve_attempts`, and
`crossed_path_indices`. The problem is purely that it is **transient** and **not reflected
per path**:

- nothing in the records store carries it, so a recalled/replayed solve loses it entirely;
- `metadata[i].endgame_success_code` for `i in crossed_path_indices` is still `Success`.

In a cellular-decomposition port built on this library, endpoints produced by flagged
paths were consumed as clean geometry — recorded, deduplicated into a vertex set, and
used to build cells. The flag was in the terminal only. Reconstructing which paths were
affected required cross-referencing `crossed_path_indices` by hand, after the fact, in a
process that had already thrown the solver away.

## Requested

1. **Persist the `MidpathCheckReport` in the records store** alongside the solve's config
   and per-path metadata, so it survives recall/replay and can be queried later.
2. **Give flagged paths a distinct success code** — e.g. `CrossedPathUnresolved` — or,
   if changing the code is too disruptive, add an explicit per-path boolean on
   `SolutionMetaData` (`crossing_unresolved`). The key property is that a consumer
   reading only per-path metadata must be able to tell a suspect endpoint from a clean
   one, without knowing to ask a separate object.

Point 2 matters more than point 1: `Success` on a solution the library has just warned
[...]

PR #428, opened 2026-09-16
feat(zero_dim): per-path singular values and an unresolved-crossing flag in the solution metadata (#409, #365)

ofloveandhate:
Two additive fields on `SolutionMetaData`, each the per-path form of something the solver already
knew and threw away or only printed.

## `singular_values` -- #409, the "expose the right raw quantity" half

`condition_number` was already σ_max/σ_min from a `JacobiSVD` of the homogenized, patched target
Jacobian at the endpoint, at the endpoint's own precision; the spectrum was discarded. It is now
kept (`Vec<double>`, largest first), carried through the thread/MPI `FullPathResult`, archived in
each path record (`singular_values`, an array of exact double texts) and restored on recall (old
records without the field simply load without it).

**Deliberately no `numerical_rank` verdict**, although #409 sketched one. A rank read needs a
tolerance, and the downstream decomposition that motivated #409 later found that every
magnitude-based rank read it tried -- including the validated one -- misjudged a genuine regular
root whose σ_min/σ_max was 6e-8 (b2_experiments, 2026-09-14); it replaced the rank read with a
Newton convergence-order test. Publishing the spectrum lets a caller do that, or any other rule, at a
tolerance of its choosing; a verdict baked in here would be wrong for someone. `is_singular` is
unchanged.

## `crossing_unresolved` -- #365, point 2

When the endgame-boundary crossing check still finds crossings after the re-track rounds, the
solver prints "the affected solutions may be wrong" and leaves every success code reading
`Success`. Now each path still flagged after the last check carries `crossing_unresolved = true`
in its metadata -- set at one point (`FlagUnresolvedCrossings`) from both the serial flow and the
distributed manager, after the resolve loop, so a successful re-run (which installs fresh
metadata) never inherits it. It is not archived: recall replays the boundary data and re-runs the
crossing check, so the flag is recomputed exactly. It is also a column of `to_dataframe()`.

Point 1 of #365 (persisting the whole `MidpathCheckReport` in the records) is left for the records
work.

## Tests (C++, `zero_dim.cpp`)

- `unresolved_crossing_is_flagged_on_the_affected_paths`: the planted cubic crossing from the
  Python suite, ported -- Euler, step 1/10, loose tolerance -- so the curved root jumps onto the
  flat one at t = 9/10. With re-tracking off the report fails and exactly the paths in
  `crossed_path_indices` carry the flag; with two re-track rounds, if the report passes nobody does.
- `endpoint_singular_values_are_the_spectrum_behind_the_condition_number`: two simple roots give
  two singular values, descending, positive, with `condition_number` equal to their ratio and
  modest; a double root collapses σ_min below 1e-3 σ_max and is classified singular.

Bindings expose both fields with docstrings; Python interface tests follow with the bindings
rebuild.


ofloveandhate (commented):
i really like that this essentially exposes the underlying reasoning behind what was previously collapsed to a single true/false, or was forgotten/only printed.  this enables more interesting workflows.  solid.  i particularly like the unit test that explicitly constructs paths that will cross, and checks that they are left labeled correctly.

Issue #392, opened 2026-08-27
Solver config surface: final_tolerance has two owners (endgame setting silently reverted), and config_names() returns names get_config() refuses


ofloveandhate:
## Summary

Two related interface defects in the solver configuration surface. Both are silent — no exception, no warning — and together they make it easy to believe you have set a tolerance when you have not.

1. **`final_tolerance` has two owners, and the loser is silently discarded.** It exists on `TolerancesConfig` (solver level) *and* on `EndgameConfig`. `solve()` pushes the solver's `TolerancesConfig` down into the endgame, clobbering anything set directly on the endgame. Setting it via `get_endgame().set_endgame_settings(...)` therefore has **no effect and raises nothing**.

2. **`config_names()` returns strings that `get_config()` refuses.** `get_config` wants a *type*; `config_names()` advertises names it will not accept.

## Reproducer — defect 1

```python
import bertini, bertini.nag_algorithm as na
# ... build homotopy HOM, start points START, target TGT ...

s = na.HomotopySolver(HOM, START, TGT, mptype='adaptive', endgame='powerseries')
eg = s.get_endgame()
cfg = eg.get_endgame_settings()
cfg.update(final_tolerance='1e-8')
eg.set_endgame_settings(cfg)
print(eg.get_endgame_settings().final_tolerance)          # 1e-08   <- it took
s.solve()
print(s.get_endgame().get_endgame_settings().final_tolerance)  # 1e-11  <- SILENTLY REVERTED
```

versus the flat route, which works:

```python
s.solve(final_tolerance='1e-8')
print(s.get_endgame().get_endgame_settings().final_tolerance)  # 1e-08
```

What makes this a trap rather than merely surprising is that the *other* endgame fields behave the **opposite** way. Measured on the same solver:

| field | set via endgame config | survives `solve()`? | accepted as a flat `solve()` field? |
|-------|------------------------|---------------------|-------------------------------------|
| **final_tolerance** | 1e-08 | **NO → reverted** | **accepted** |
| num_sample_points | 7 | yes | rejected (`AttributeError`) |
| max_num_refinements | 7 | yes | rejected (`AttributeError`) |
| min_track_time | 1e-120 | yes | rejected (`AttributeError`) |

So the same API surface has opposite semantics field by field, and the one that loses does so without a word.

[...]

Issue #390, opened 2026-08-26
bertini.linalg.solve / PartialPivLU: non-square mp input corrupts the heap instead of raising


ofloveandhate:
## Summary

`bertini.linalg.solve` (and the exposed `PartialPivLU`) accept **non-square** multiprecision matrices and hand them straight to `Eigen::PartialPivLU`, which requires a square invertible matrix. Depending on the shape this either **corrupts the heap** (`free(): invalid pointer`, SIGABRT — seen as a bare SIGSEGV in a larger program) or **silently returns a wrong answer**.

The `double` path is unaffected: it routes through numpy and raises `LinAlgError` correctly. So the mp and double paths of the same public API disagree about what is an error.

A crash is the worst outcome here because it is **uncatchable from Python** — a `try: ... except Exception:` around the call cannot save you, so a single bad shape kills the whole process with no traceback.

## Minimal reproducer

```python
import numpy as np, bertini, bertini.linalg as linalg
from bertini.multiprec import complex_mp as C

bertini.default_precision(30)

A = np.array([[C('2e-6'), C('-3e-6'), C('-1e-12')]])   # 1 x 3
b = np.array([C('5e-13')])                             # length 1

linalg.solve(A, b)      # -> free(): invalid pointer  (SIGABRT)
```

Control: the same call with `double` raises properly.

```python
linalg.solve(np.array([[2e-6, -3e-6, -1e-12]]), np.array([5e-13]))
# LinAlgError: Last 2 dimensions of the array must be square
```

## Observed behaviour by shape

`complex_mp`, well-conditioned entries, each case run in its own subprocess:

| shape | `linalg.solve` | `linalg.lu(A).solve(b)` | `double` (either) |
|-------|----------------|--------------------------|-------------------|
| 3 x 3 | ok             | ok                       | ok                |
| 1 x 3 | **CRASH**      | silently wrong answer     | raises `LinAlgError` |
| 3 x 1 | silently wrong answer | **CRASH**           | raises `LinAlgError` |
| 2 x 4 | **CRASH**      | silently wrong answer     | raises `LinAlgError` |

`lstsq`, `qr` and `svd` handle all of these shapes correctly (they are backed by
`ColPivHouseholderQR` / `JacobiSVD`, which are defined for rectangular input).

## Cause

`python_bindings/src/linalg_export.cpp`:

```cpp
template<typename T>
Vec<T> SolveLinearSystem(Mat<T> const& A, Vec<T> const& b)
{
[...]

Issue #386, opened 2026-07-27
an overdetermined system causes crashes


ofloveandhate:
homotopy solver plus overdetermined equals hard crash from python.  bad!

ofloveandhate:
Same crash as #383 (a homotopy with more functions than variables), which carries the reproduction. Tracking it there; closing this one as a duplicate. The fix planned for 3.5 checks the function and variable counts at the Python boundary and raises, instead of letting the C++ side abort the process.

Issue #383, opened 2026-07-27
Hard crash when setting up a homotopy with more functions then variables.


ElderStack:
I set up the homotopy wrong however it hard crashes instead of throwing an error. Below is the homotopy and it crashes when I attempt to solve it.

```
mover = bertini.HomotopySolver(hom, startpoints, target)
results = mover.solve() #<- crash here
```

This is the homotopy:
```
1 variable group:
  group 0: x y z 

5 functions:
  f_0 = (x^2+y^2+z^2-1)*(y-x^2)*(x-2)
  f_1 = (x^2+y^2+z^2-1)*(z-x^3)*(y-2)
  f_2 = (x^2+y^2+z^2-1)*(z-x^3)*(y-x^2)*(z-2)
  f_3..f_4 = (1-t)*A + ((1793251369130129/5804064888416,4442615913964055/1976558544054797)*t)*B   (blend of 2 systems)

path variable: t
```

ofloveandhate:
Fixed on develop by #429: a homotopy with more functions than variables is refused with a ValueError when the HomotopySolver is built, instead of aborting the process inside the tracker. Ships in 3.5.0.

PR #429, opened 2026-09-16
fix(python): caller errors at the Python boundary are Python exceptions, never an abort

### LINKED ISSUE #369: start-point/system size mismatch aborts the process (std::terminate) instead of raising a Python error
## Summary

Passing a start point whose length does not match the tracked system's variable count to `HomotopySolver` (via `bertini.nag_algorithm`) does not raise a Python exception — it throws a C++ `std::runtime_error` that escapes to `std::terminate` and **aborts the whole process** (core dump). Under OpenMP the abort prints `terminate called recursively` from multiple threads.

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  start point size must match the number of variables in the system to be tracked
terminate called recursively
Aborted (core dumped)
```

## Repro sketch

```python
import bertini, numpy as np
import bertini.nag_algorithm as na
# any moving homotopy over a 3-variable system...
solver = na.HomotopySolver(homotopy, [eight_var_point], target,
                           mptype='adaptive', endgame='powerseries')
solver.solve()   # process dies here; no Python traceback possible
```

Hit in practice when a caller accidentally supplied an 8-coordinate point (from a doubled-chart lift store) to a 3-variable tracker: instead of a `ValueError` naming the mismatch, the interpreter (and everything in flight — long decompositions included) is killed.

## Expected

A Python `ValueError` at solver construction (or at `solve()`), e.g. `start point 0 has 8 coordinates; the tracked system has 3 variables`. The size check evidently exists (the `what()` message is descriptive) — it just throws from a context where the exception cannot propagate to Python (likely inside an OMP parallel region or a noexcept boundary), so it must be validated **up front, before tracking begins**, on the Python-facing side of the call.

## Doctrine

Per the standing rule that no C++ converter/type error should reach Python raw (and certainly should never abort the process): every start point's length should be validated in the bindings layer with a friendly message.

ofloveandhate:
Closes #369, #383, #390, #367, #372, #392 (and the duplicates already folded into them, #386 and
#381).

One doctrine, six sites: **a caller error must be a Python exception raised at the boundary --
never a `std::terminate`, never heap corruption, never a raw Boost.Python argument error.**

## What each site did, and does now

| site | before | now |
|---|---|---|
| `HomotopySolver(H, start_points, target)` with a start point of the wrong length | the tracker threw from inside the tracking loop, where nothing can propagate: the process aborted with no traceback, taking every computation in flight with it (#369) | `ValueError` at construction naming the point, its length and the variable count |
| the same with a homotopy that has more functions than variables | the same abort (#383, #386) | `ValueError`: tracking needs a square system; randomize down or cut with a slice |
| `bertini.linalg.solve(A, b)`, A not square or b the wrong length, multiprecision | Eigen's LU preconditions are only asserted in debug builds; a release build corrupted the heap (#390) | `numpy.linalg.LinAlgError` from the Python wrapper (as the double path already raised), and the native `solve` refuses too (`invalid_argument` -> `ValueError`) |
| `System.eval([0.3, 0.7])` | a Boost.Python argument error listing C++ overloads (#367) | a list of ordinary numbers evaluates in double; a list with any multiprecision entry in multiple precision |
| `System.add(slice)`, `moving_homotopy(fixed, slice_a, slice_b)` | `TypeError` / argument error (#372, #381) although `Slice.add_to` and `Slice.as_system` existed | accepted: a slice's linear forms become a linear-forms block; a slice is taken as the system of its forms |
| `solver.get_config('tolerances')` | `KeyError`, although `config_names()` had just listed that name (#392, defect 2) | the name resolves through the same table `configure()` uses; the class form is unchanged |
| `final_tolerance` set on the endgame, then `solve()` | silently reverted at every solve by the solver's copy, so a whole parameter sweep through the endgame config did nothing (#392, defect 1) | the solver's value flows into the endgame at setup and whenever the solver's value *changes* (`PushFinalToleranceToEndgame`); whichever was set last wins |

The `final_tolerance` change is the only core change: `PreSolveSetup` no longer pushes
unconditionally. Classic-input `FinalTol` still flows (the blackbox sets the solver's
`TolerancesConfig`, which pushes), and the constructor's `DefaultSetup` does the first push, so a
value set on the endgame before the first solve survives it -- the issue's exact reproduction.

Not touched: the eigenpy decomposition visitors exposed beside `solve` (`PartialPivLU`, `QR`,
`SVD`) still carry only Eigen's debug asserts; that audit is a follow-up.

## Tests

- C++ (`zero_dim.cpp`): `final_tolerance_set_on_the_endgame_survives_a_solve` -- setup pushes the
  solver default; a value set on the endgame survives `Solve()`; a later solver-level `Set` wins.
- Python (`python/test/zero_dim/seam_validation_test.py`, interface): every row of the table --
  wrong-length start point, overdetermined homotopy, a well-shaped one still solving, non-square
  and mis-sized `linalg.solve` for double and multiprecision plus the native guard, `eval` with
  lists in both precisions, `System.add(slice)` and `moving_homotopy` with slices solving to the
  right roots, `get_config` by every listed name, and the `final_tolerance` precedence.


ofloveandhate (commented):
turns crashes into exceptions, that's a good thing.  i found some of the code a little cryptic, in that i had to read it twice.  but on the second reading, i understood it.  so i am accepting it.

Issue #421, opened 2026-09-15
Adopt the MaRDI file format (.mrdi, JSON) for interchange and persistence (target 3.6)


ofloveandhate:
## Proposal

Adopt the MaRDI file format (`.mrdi`: JSON with a published schema, used by OSCAR, polymake, GAP
and Singular to exchange typed mathematical objects) as bertini2's format for interchange and,
eventually, for the records archive.

## Why

- bertini2 should not be captive to the Bertini 1 classic input format for anything except
  talking to Bertini 1. That format cannot express what bertini2 does -- variable groups with
  types, homogenizing variables, patches, path variables, blend and randomization blocks,
  non-polynomial functions -- and its number syntax is limited.
- A schema-backed JSON envelope with exact coefficients and references for shared sub-objects is
  what the archive needs anyway. Using the one the computer-algebra systems have converged on
  means bertini2 systems can be opened there, and theirs here.

## Plan (target: 3.6)

3.5 ships a reader for bertini2's own canonical encoding, which makes every existing records
directory reloadable and provides the System builder that the MaRDI reader will reuse. Then:

1. Study the schema and its extension conventions (namespaces, type names, references).
2. Decide the mapping: standard types where they apply (polynomials over the rationals,
   matrices); a `bertini2` namespace for the rest.
3. Settle how arbitrary-precision real and complex constants are carried, with their precision.
4. Writer and reader in the core, bound to Python. Acceptance: the reloaded system's content
   digest equals the original's, and the files validate against the published schema.
5. Records directory format `b2rec/2`, with a migration of `b2rec/1` stores through the 3.5 reader.
6. An ADR.

MathJSON was considered and set aside: it models expressions only, so everything else would
still need a bertini2 envelope, and its users are web and CAS front ends rather than the systems
above.

PR #430, opened 2026-09-16
feat(records): a reader for the canonical encoding -- archived systems reload by digest

ofloveandhate:
The records archive stores every system and homotopy a solve touched as its exact canonical
encoding -- the text the content digest is the SHA-256 of.  Until now nothing could read that
text back.  This adds the reader, so an archived system comes back to life with its digest as
the proof it is the same object.  Classic (Bertini 1) text plays no part in persistence: it is
an input format and cannot express block structure.

## What is added

**C++**

- `node::DecodeCanonical` / `node::DecodeCanonicalTree` (`function_tree/canonical_decoding.hpp`):
  a recursive-descent reader for the node grammar -- every leaf kind (variables, integers,
  rationals, complex literals **at their stored precision**, pi, e, differentials), every
  operator kind, netstring names, and back-references resolved through the same
  first-encounter numbering the encoder uses.  Nodes are rebuilt through their ordinary `Make`
  factories, so the result is interned: decoding an encoding of a live tree gives back the very
  same object.
- `System::FromCanonicalEncoding(text)`: the inverse of `CanonicalEncodingText()`, section by
  section -- variable groups with their types and time order, path variable, parameters, every
  block kind with its exact coefficients (polynomial, linear forms, products of linears,
  randomization with its operand system, blend with its operand systems), the patch, and the
  pre-homogenization snapshot.  Operand systems decode recursively with the shared
  back-reference table, as they were encoded.
- `records::LoadSystem(directory, digest)` (`records/load_system.hpp`): fetch the definition
  document, decode its `encoding`, **check the rebuilt digest equals the one it was filed
  under**, intern.  A document that does not reproduce its digest is refused.
- Small restoration constructors the reader needs: `Patch(sizes, coefficients)`,
  `LinearFormsBlock(num_vars, coefficients, homogenized)`,
  `RandomizationBlock::RestoreHomogenization(hom_vars)` (marks the block homogenized without
  re-homogenizing an operand that already is).

**Refusals** (all `std::runtime_error`, naming the byte offset and a snippet): an encoding of
a version this build does not write (`b2sysenc/2`), one written under other session
canonicalization settings (monomial order, canonicalize-by-default, power-fold -- they are in
the header because they shape the trees, so decoding under other settings would not reproduce
the digest), truncated or trailing text, unknown node or block kinds, dangling
back-references.

**Python**

- `System.canonical_encoding()` and the static `System.from_canonical(text)`.
- `bertini.records.load_system(digest, directory=None)`; `OutputDirectory.get_definition(id)`
  and `has_definition(id)`.

## Tests

- `core/test/classes/canonical_decoding_test.cpp`: every node kind round-trips to the same
  object and the same text; names with any bytes (emoji, spaces, parentheses); complex literals
  keep precision and digits, including a product mixing precisions; back-references within a
  tree and across roots sharing one context.  Systems: plain, rational and transcendental
  coefficients, two affine groups, a projective group, homogenized, homogenized and patched, a
  total-degree homotopy with a blend block, randomized (affine and homogenized), linear-forms
  and products-of-linears blocks with mixed-precision coefficients -- each checked by text
  equality, digest equality, `IsSame`, and agreement of evaluation at a point.  Decoded systems
  share the live interned variables.  Every refusal above.
- `core/test/nag_algorithms/zero_dim_records.cpp`: after a recording solve, the target system
  and the homotopy reload from the directory by digest; loading interns (same digest twice is
  one object); a document filed under a digest it does not reproduce is refused.
- `python/test/records/records_test.py`: the interface -- `load_system` returns a system that
  `is_same` the one solved, interned; `from_canonical` round-trips; the two refusals.

## Not in this PR

MaRDI's `.mrdi` JSON as the interchange format (#421) -- a 3.6 target whose reader reuses
this one's System builder.  This PR makes the archive reloadable with the format it already
has.


ofloveandhate:
Great, solves an open end.  Glad I had an AI to write it for me -- this kind of boilerplate coding isn't rewarding to write, for me any more.  I am unashamed.  I'm writing free software, and Claude is SO helpful for this.  

I do note, this is not the final form -- moving to MaRDI, or another interoperable format, is actually the goal.  But this fills a round-trip hole. so i'm taking it.

PR #431, opened 2026-09-16
fix(system): a slice with more linear forms than variables is refused

### LINKED ISSUE #380: what happens if i try to generate a slice with more slices than variables?
because the slice coefficient generating algorithm uses an orthonormalization algorithm, if a slice is requested with more rows than there are variables, i do not know what will happen.

ofloveandhate:
Closes #380.

The question was: what happens when a slice is asked for with more linear forms than
variables?  The answer was: nothing good, quietly.  The orthonormalization that draws the
coefficient rows cannot make more independent rows than there are columns, so the extra rows
came back dependent, and a slice that cannot cut anything new was handed over as if it were
fine.

Now every random factory (`RandomComplex`, `RandomReal`, homogeneous, non-orthogonal, and the
through-point forms) refuses with `std::invalid_argument` (a `ValueError` in Python) that
names both counts.  A homogeneous slice through a point is held one tighter: its forms live in
the point's orthogonal complement, which has one dimension fewer than there are variables, so
`dim` must be at most `num_vars - 1` there.  As many forms as variables is still allowed.

Two tests from 2016 over-sliced on purpose: `slice_basic_crazy_overslice` pinned only that a
6-on-3 request came back with a 6x3 shape, and `head_tail_rows_subsetting` used four forms on
three variables as a convenience.  The first is now the refusal check; the second has a
fourth variable.  No tutorial, library code, or other test over-slices.

## Tests

- C++ (`slice_test.cpp`): `more_forms_than_variables_is_refused` -- every factory and the
  homogeneous-through-point bound refuse; the boundary cases (as many forms as variables, one
  fewer through a point) still work.
- Python (`python/test/nid/slice_shape_test.py`, interface): the refusal reaches Python as a
  `ValueError` that says "more forms than variables"; a full slice still works.


ofloveandhate (commented):
geometry, ah geometry.  excellent to close this opportunity for nonsense.

PR #432, opened 2026-09-16
test(function_tree): the MultiDegree scaling test takes the minimum of interleaved trials

ofloveandhate:
Follow-up to #418.  The scaling test `multidegree_cost_follows_the_dag_not_the_expansion` times
`MultiDegree` on a 12-level and an 18-level expression and asserts the ratio is under 8 (the
memoized walk is ~1.5; the unmemoized one is 64).  It timed each size once.  On #430's rerun
a shared macOS runner stalled during the 18-level timing and the ratio read 11.5, failing a
PR whose code had nothing to do with it.

The intrinsic cost is the **minimum** over several trials -- the only statistic a stall cannot
inflate.  The test now takes the minimum of seven trials per size, interleaved (12, 18, 12,
18, ...) so a slow stretch of the machine cannot land on all of one size's trials.  The limit
stays at 8.  No library code changes.


ofloveandhate:
okie dokie.  testing workarounds for CI unreliability, seems fine to me.

Issue #361, opened 2026-07-16
add more events to the various algorithms


ofloveandhate:
it was impossible to render just the real-axis time plot of paths from cauchy, because we couldn't really tell which time points were actually on the path, and which were partway-round points during cauchy loops (which have real time, but aren't actually on the path).

i need to add more event types, so observers can do a better job of reporting interesting things.

ofloveandhate:
Status, since a chunk of this landed in v3.5 and the issue does not say so.

**Landed (#433):** the endgames announce the sample points they compute. `ComputedSamplePoint` carries the point and its time, from both the power series and the Cauchy endgame, and a `SampleSequenceCollector` observer (C++ and Python) gathers them -- the sequence of points on the path at shrinking times, kept apart from the Cauchy loop's circle points and from the successive approximations of the root, with run boundaries so one collector on a solver's endgame can tell one path from the next. Every endgame run also announces `Initializing`, and an adaptive endgame that needs higher precision before its first approximation announces `SamplesRecomputedAtHigherPrecision` so a collector drops the superseded samples as the endgame does.

That is what lets a caller watch a quantity at the root -- a Jacobian's singular values, say -- as a function of distance to it, rather than judging it at one point.

**Still open:** the Cauchy loop/shrink phase events, which would let a consumer filter loop points exactly rather than by inference, and whatever else an event audit turns up for the trackers and the solver. The observable-event schema question (what an event carries, uniformly) is unanswered.

Leaving open for that.

PR #433, opened 2026-09-16
feat(endgames): every computed sample is announced; a SampleSequenceCollector gathers the approach

ofloveandhate:
Advances #361.  This is the last piece of the port's integration branch that develop did not
have: the endgames announce every sample point they compute, and an observer collects them
as the sequence of points approaching the root.

## What is added

- **`ComputedSamplePoint` event**, from both endgame flavors, wherever a new path sample
  becomes final: the power series endgame's time advance (after the new sample is refined)
  and the Cauchy endgame's rotation onto its power-series window.  The point and its time
  ride on the event, so an observer never has to know which flavor emitted it.  The initial
  sample window is announced too, starting with the endgame's own input point at the
  boundary time -- without that the sequence started at `boundary * sample_factor^(n-1)`
  and threw away most of the span before anyone looked at it.
- **`Initializing`** is now emitted by every endgame run (it was declared and handled by
  `GoryDetailLogger`, but emitted by nothing), so a collector attached to a solver's endgame
  can tell one path from the next.
- **`SamplesRecomputedAtHigherPrecision`** (new): an adaptive endgame that needs a higher
  precision before it has its first approximation recomputes its sample window at the new
  precision, and the samples it announced at the lower precision are superseded.  It now
  says so (right after the matching `PrecisionChanged`), and the collector drops the
  superseded samples as the endgame does.  This is what the earlier flaky test was seeing:
  on the adaptive power series endgame the sequence read `0.1, 0.05, 0.025, 0.1, ...` -- the
  window at double precision, then the same window again in multiple precision.  The
  collector's times now march toward the target within every run.  A precision increase
  later in the run, once approximations exist, keeps the samples and widens them in place;
  that one announces only `PrecisionChanged`.  The collector counts every increase
  (`num_precision_increases`).
- **`SampleSequenceCollector`** (C++ and Python, `bertini.endgame.observers.<flavor>`):
  three buckets kept apart -- `path_samples`/`path_times` (the sequence), `circle_samples`
  (Cauchy loop points at constant |t|, not part of the sequence), `approximations` with
  their errors and cycle numbers -- plus `advance_times`, per-run start indices, `num_runs`,
  `num_samples`, `num_precision_increases`, `clear()`.

The vocabulary is **sequence** throughout (the integration branch said "ladder"; renamed on
the way in).

## Tests

- C++: `sample_sequence_collector_separates_path_from_circle` (Cauchy) and
  `sample_sequence_collector_serves_power_series_too` (power series), each across the fixed
  and adaptive endgames: buckets consistent, times strictly shrinking toward the target,
  the sequence starts at the boundary point (so a recomputation provably dropped the
  superseded samples), the approach is real, exactly one run observed, `Clear()` empties
  everything.
- Python (`python/test/tracking/sample_sequence_test.py`, interface): the collector on the
  adaptive power series endgame and on the fixed Cauchy endgame, reading every bucket.


ofloveandhate (commented):
This is correct.  It's actively used in my-in progress port of bertini_real to use b2.  i needed this to be able to compute the rank of the jacobian, on the sequence of endgame samples that leads to the endpoint.

PR #434, opened 2026-09-16
fix(classic-io): comments are removed in C++ before any classic input is parsed (#407)

### LINKED ISSUE #407: Classic parser rejects any commented input: no comment skip rule (C++), and the INPUT/CONFIG finder scans raw text (Python)
## Summary

`bertini.parse.system` cannot read a commented classic Bertini input file. Two
independent defects, one in the C++ parser and one in the Python wrapper, and either alone
is enough to reject a file. Since classic syntax is b2's Bertini 1 **input/compat** path,
and real Bertini 1 files are commented as a matter of course, this makes the compat path
unusable for anything but machine-generated bodies.

Found while writing an exact classic input file by hand for a slice of the Whitney
umbrella; the file is valid Bertini 1 and b2 rejects it.

## 1. The classic parser skips no comments at all (C++)

`core/include/bertini2/io/parsing/system_rules.hpp` (and friends) have no skip rule for
comments, so any comment line is a parse error. Bertini 1 uses `%`.

```python
from bertini.parse import _native_system   # the raw C++ entry point
_native_system('variable_group x, y;\nfunction f;\nf = x^2 - y;\n')      # OK
_native_system('variable_group x, y;\nfunction f;\n% a note\nf = x^2 - y;\n')
# RuntimeError: [SystemParser] could not parse a system from the input.
#   Expected classic declarations (variable_group / function / definitions), ...
```

Measured, every comment style is rejected, including a trailing one on an otherwise
valid line:

| comment | result |
|---|---|
| `% note` (Bertini 1) | REJECTED |
| `// note` | REJECTED |
| `# note` | REJECTED |
| `f = x^2 - y; % note` | REJECTED |

The fix is a skip parser: Qi's `qi::space` can be replaced with a skipper that also eats
`'%' >> *(qi::char_ - qi::eol) >> qi::eol`.

## 2. The section finder scans the raw text, so a comment can hijack it (Python)

`python/bertini/parse/__init__.py::_input_body` locates the `CONFIG` and `INPUT` sections
with a regex over the whole string, comments included:

```python
m = _re.search(r'\bINPUT\b(.*?)\bEND\s*;', s, flags=_re.IGNORECASE | _re.DOTALL)
```

So a comment containing the word `input` before the real section becomes the match. A
header comment as ordinary as this is enough:

```
% Bertini input: critslice 8 of the Whitney umbrella
...
CONFIG
  TrackType: 1;
END;

INPUT
  variable_group x, y, z;
  ...
END;
```

The reported error quotes the comment, which is a good clue to the cause once you know to
look:

```
RuntimeError: [SystemParser] could not parse a system from the input. ...
  Input began: ": critslice 8 of the Whitney umbrella
%
% ..."
```

`\bCONFIG\b` has the same exposure. The fix is to strip comments before locating the
sections (or to anchor the search to a statement position).

## Not the cause

Bisected, so these are ruled out and need no attention:

| construct | result |
|---|---|
| exact rational constant, `a = 3/7;` | OK |
| 46-digit-over-45-digit rational constant | OK |
| a function named `slice` | OK |
| several separate `constant` declarations | OK |

## Suggested test

A round-trip: emit a system with the classic writer, add a `%` comment line and a header
comment containing the word `input`, and parse it back. Both defects fail that test, and
neither fails any current one.

ofloveandhate:
Closes #407.

Bertini 1 comments (`%` to the end of the line) may sit anywhere in a classic input: whole
lines, or the tail of a declaration.  The grammars have no skip rule for them, so a commented
input failed to parse through `System(text)` and `parse.system` (the CLI's file path already
stripped comments before splitting).  The Python side had its own trap: `parse.system`
scanned the raw text for the `INPUT ... END;` section, so a comment that mentioned `INPUT` or
`END;` moved the goalposts.

**C++ owns it now.**  `StripClassicComments` (beside the BOM and file-wrapper strippers in
`qi_files.hpp`) removes everything from a `%` to the end of its line, keeping the line breaks
so error positions still name the right line.  It runs at every parse entry point *before*
the file wrappers are unwrapped, so a commented-out `INPUT` or `END;` means nothing.  Bertini 1
has no string literals, so a `%` is always a comment marker.

The Python `_input_body` scanner is deleted -- it was already redundant since #398 wired the
wrapper stripping into the bound parser.  `parse.system` keeps only the `(re,im)` complex
literal rewrite, which goes away with the emitter work (#417's second half).

## Tests

- C++ (`classic_parsing_test.cpp`): a commented body and a commented full file -- title line,
  trailing comments, a commented-out function, a comment containing `function END; INPUT`, and
  a final comment with no newline after it -- both parse to a system `IsSame` as the clean one.
- Python (`parser_test.py`, interface): the same text through `parse.system`, bare and wrapped.


ofloveandhate (commented):
I would have preferred this to not just be using std::string but something more formal.  But this is functional, and I don't have infinite time, so this is good enough.

Issue #420, opened 2026-09-15
Recall can return results a fresh run would not produce: the ask identity has no slot for the algorithm's own identity


ofloveandhate:
## What happens

A solve is recalled from the records directory when its ask matches a recorded one. The ask is
the target system's digest, the homotopy's digest, the settings digest, the seed, and the tracker
and endgame kinds. None of those identify the *code* that produced the recorded paths. So after
any change in behaviour with the same inputs -- a tracker or endgame fix, a change to a config
default that is not part of the settings digest (`SharpeningConfig` and `MidPathConfig` are the
known gaps), or any algorithmic change -- a recall returns what the old code computed, silently,
where a fresh run would compute something else.

Concretely: a records directory written before #411/#416 (a path that could fail steps forever)
or before the 3.4.0 endgame fixes is recalled unchanged by a build that contains those fixes.

## Why it matters

Recall is meant to be memoization: the same answer, cheaper. Without the algorithm's identity in
the ask it is an older answer, cheaper, and nothing in the directory says which.

## What is not the problem

Changing a config default that *is* in the settings digest is handled correctly today: the digest
encodes values, so an ask that used the default changes identity and is recomputed.

## Possible directions (not decided)

- Put an algorithm identity in the ask: at minimum the bertini2 version; better, a per-algorithm
  version token that is bumped when behaviour changes, like the encoding version tokens.
- Add the missing configs to the settings digest.
- Decide what a recall of a record from an older version should do: refuse, warn, or recompute.

Until this is settled, a records directory is a cache of what the version that wrote it
computed, and the cross-version "records are read forever" promise is withdrawn. (3.5 will also
bump the system encoding version, so 3.4 records will not be recalled by 3.5.)

ofloveandhate:
## The ask is keyed on the wrong axis

The problem above is real but it is half of a larger one. The missing algorithm identity makes the ask too *coarse*. The settings digest, at the same time, makes it too *fine*. Both follow from the same mistake: the ask keys on how a run was configured rather than on what was asked for.

## What the ask is today

Built in `nag_algorithms/zero_dim_solve.hpp` (`RecordsAsk`): `op`, the tracker kind, the endgame kind, the target system digest, the homotopy digest, the start identity when there is external start data, the settings digest, and the seed. The only thing in there that refers to the algorithm is its *name*. Everything else that discriminates one ask from another is inputs plus settings.

## Settings are two different kinds of thing wearing one coat

[...]

PR #435, opened 2026-09-16
change(endgames)!: the default endgame is power series, matching Bertini 1 (ADR-0058)

ofloveandhate:
**The default endgame is the power series endgame**, matching Bertini 1's documented default
(`EndgameNum: 1`).  ADR-0058 carries the decision and the measurements; this PR is the flip.

## What changes

Every choice point that used to default to Cauchy now defaults to power series:

- `EndgameChoiceConfig::endgame` (so the CLI, and a classic input without `endgamenum`);
- the Python factories `ZeroDimSolver`, `HomotopySolver`, `user_homotopy`, `parameter_sweep`,
  and `bertini.solve`.

Cauchy remains fully available by explicit request (`endgamenum: 2`, `endgame='cauchy'`).

## Why

Measured on a regeneration-cascade workload with library defaults: Cauchy 190 s, power series
2.5 s (equal to Bertini 1's wall time on the same system and seed).  Slowly diverging paths are
a Cauchy-specific pathology -- its at-infinity truncation is an absolute norm threshold reached
only after many decades of t, and its security valve is gated to an operating zone a slow
diverger never enters.  Bertini 1's own Cauchy endgame is ~11x slower than its power series
endgame on cyclic-6 while both find all 156 solutions.  Cauchy is the specialist tool, not the
default.

## Identity

The endgame kind is part of the ask, so a solve that relied on the default is a different
computation from before and its records are new asks rather than recalls.  That is correct.
No encoding-version bump: the canonical encoders are unchanged.  The config keyspace test
pins its `EndgameChoiceConfig` recipe to Cauchy so it keeps measuring the encoder rather
than the default; the golden digest fixture and the version registry are untouched (the
test proves it).

## Tests and docs

- `zero_dim_factory_test.py`: the defaults resolve to the power series solver classes.
- The `precision_models` tutorial (doctested) names the power series classes.
- The whole C++ suite, the full pytest suite and the tutorial doctest gate pass with the flip.
- ADR-0058 added, with its index row.

The parked stash from July also carried a "defaults are not identity" rewrite of the config
keyspace; that part is deliberately NOT here (see #420 for the deeper question it belongs to).


ofloveandhate (commented):
correct.  some places where i would love to automatically infer defaults from the core into the python bindings, rather than hardcoding them in a way that permits drift.  but this is ok for now.  i will not get everything perfect as i do this -- just more perfect over time, never reaching my goal.

PR #437, opened 2026-09-17
chore: drop the per-file author lists from every source header

ofloveandhate:
Removes the per-file "individual authors of this file include:" comment block from every
source file that still carried one -- 243 files under `core/`, `python_bindings/` and
`python/`.  The GPL header stays; so does whatever followed the block (a `\file` doc
comment, a one-line file description, includes, or a module docstring), with exactly one
blank line between the license and it.

The lists were stale (names, affiliations and semesters from 2015 on, most files never
updated), and `git log` is the record of who wrote what.  Nothing but names, affiliations,
dates and empty comment lines was removed: every distinct removed line was reviewed by eye.
The one block that also carried a third-party attribution (`enable_permuted_arguments.hpp`,
the Stack Overflow post its templates are based on) keeps that attribution in the file's
`\file` doc block, which already had it.

Comments only: the whole C++ suite, the full Python suite and both doc lints pass unchanged.


ofloveandhate:
i thought they were a good idea when i wrote them initially well over a decade ago...

PR #449, opened 2026-09-18
feat: Ctrl-C stops a solve, cleanly, and the solver is left inspectable

### LINKED ISSUE #400: wishlist: bound what a single path may cost (no mechanism today to stop a long-running path)
## Wishlist: a way to bound what a single path may cost

There is currently no mechanism to stop a path that is taking too long. The only budget a
tracker carries is `SteppingConfig::max_num_steps` (default `1e5`, → `SuccessCode::MaxNumStepsTaken`),
and a step count is a poor proxy for cost: a step at 90 digits costs orders of magnitude more
than one at 16, so the same budget bounds wildly different amounts of work depending on where
adaptive precision took the path.

## Field evidence

From a cellular-decomposition consumer, in a comment written to explain why a correctness
improvement had to be left *off by default*:

> moving a fiber endpoint onto its exact tangency makes that boundary curve's interslice
> tracking explode (measured: seed 204, one boundary curve **2s → 130s+**; not RNG, not
> spurious)

A single path costing 130 seconds is bad on its own, but the second-order effect is worse:
because there is no way to bound it, a *correct* refinement had to be disabled to keep runs
finishing. An unbounded path cost turns "this makes the geometry right" into "this makes the
run unusable", and the correctness work loses.

That is the shape of the request — not "make paths faster", but "let a caller say how much a
path is allowed to cost, and get an honest failure code when it exceeds that."

## What would help

A per-path budget that a caller can set and that produces an ordinary non-success code
(`SuccessCode::` — a new one, or a reuse), so the existing retry/rescue machinery treats it
like any other honest failure rather than a hang.

**Please prefer a WORK budget over wall-clock.** Wall-clock makes a run non-reproducible: the
same recorded solve would truncate on a loaded machine and complete on an idle one, so a
replay would not reproduce the original. That directly undercuts the records contract
(`b2rec/1`, ensure-answered recall, ADR-0042..0047) — what got computed should not depend on
machine load. A deterministic budget truncates in the same place every time.

Candidate work measures, roughly in order of how well they track real cost:

1. **precision-weighted steps** — e.g. accumulate `steps × digits` (or `digits²`), which
   bounds actual arithmetic rather than iteration count;
2. **a step budget that is actually consulted per-path by callers** — `max_num_steps` exists
   but at `1e5` it is effectively no bound; the useful thing may be documentation plus a
   sane default, since a consumer already lowers it to 2000 for rescue passes;
3. wall-clock, but only as an opt-in *safety valve* on top of a work budget, clearly marked
   as making runs non-reproducible.

## Diagnosing which one bites

`SolutionMetaData` already reports `path_time_seconds`, `max_precision_used`,
`precision_changed` and `time_of_first_prec_increase` per path, so it is already possible to
tell whether a slow path is *many* steps or *expensive* ones. If a 130s path shows
`max_precision_used` at 90+, the cost driver is precision escalation rather than step count,
and a plain step cap would not have helped — which is the main argument for a
precision-weighted budget over a step count.

Version: `3.5.0.dev0` (develop @ 8f0cd0f1).

ofloveandhate:
Press Ctrl-C during a solve and, until now, nothing happened until the solve finished. In a notebook that meant a trapped kernel and killing the process as the only exit. This makes Ctrl-C stop the solve at its next step, cleanly, with the solver left exactly as the stop found it.

## Why the kernel trapped

`solve()` releases the interpreter lock and runs on the calling thread. CPython's C-level handler notes the SIGINT immediately, but `KeyboardInterrupt` is raised from the evaluation loop, and the evaluation loop is not running because the thread is inside the C++ solve. So the interrupt was delivered, recorded, and ignored until the very thing you wanted to stop had finished.

That also says what the fix is not: it is not a watcher thread, and the solve does not move off the calling thread. Moving it was investigated and rejected, since a fresh thread's multiprecision working precision can be zero and aborts inside mpfr on first use, the session guard restores thread-local random streams belonging to whichever thread runs it, and MPI is initialised funneled. None of that is needed.

## What it does

**Core.** One process-wide, lock-free stop flag: `bertini::RequestStop()`, `ClearStopRequest()`, `StopRequested()`, and an RAII `ScopedStopRequest`. Every tracking loop checks it at the top of each iteration, right where the step and precision budgets already live, and returns `SuccessCode::ExternallyTerminated`, a value that has been in the enum, produced by nothing, since long before this. Process-wide is the right scope: a keyboard interrupt is a fact about the process, and if two solves are running in threads, Ctrl-C means stop both. A bare tracker honours the request with no solver anywhere, because it is the tracker that checks.

**Solver.** A path in flight stops at its next step and reads `ExternallyTerminated`. A path not yet begun is never started and stays `NeverStarted`. The solver reports `WasStoppedEarly()` and `NumPathsNeverStarted()`. The midpath check is skipped on a stopped solve, since comparing paths against each other at the endgame boundary means nothing when an arbitrary number of them never ran. The stop is cooperative throughout: no exception crosses a worker thread, so the per-path slots are never half-written, which matters because those slots are precisely what the user then inspects.

**Bindings.** For the duration of the local solve a SIGINT handler is installed whose whole job is two atomic stores. The solve unwinds, the previous handler is restored, the request is withdrawn, and `KeyboardInterrupt` is raised. Only the C-level handler is touched, so `signal.getsignal` and any handler of the user's own come back untouched. `bertini.request_stop()` does the same thing programmatically from another thread and returns normally rather than raising; `stop_requested()` and `clear_stop_request()` round it out. `solver.was_stopped_early()` and `solver.num_paths_never_started()` are bound. Not applied to the MPI path, where a signal reaches one rank and stopping one rank of a manager-worker solve is its own problem.

**Records.** An abandoned path is recorded, with its code, because a missing record does not say why a result is absent; it is just silently missing. A path never begun is not recorded, since its absence is explained once by the solve having been cut short and thousands of rows saying "did not get to this one" would be noise rather than honesty. And recall no longer reuses an abandoned path. The rule, which will carry over to the wall-clock limit unchanged: a path record is reusable when the run that would reuse it asks no more of that path than the run that produced it. A determination always qualifies. A path somebody stopped was abandoned for a reason deliberately not in the ask, so it is re-tracked. Its record stays; it is simply not mistaken for a conclusion.

Measured before that last piece existed: an interrupted 27-path solve, solved again as the same ask, recalled 13 paths and handed back the three abandoned pre-endgame and the two abandoned in the endgame as abandoned. The fresh baseline is 27 of 27 successes. With the rule in place the second solve re-tracks those five and reads all-success.

## What you get in a session

```python
solver = bertini.ZeroDimSolver(sys)
try:
    solver.solve()
except KeyboardInterrupt:
    pass
solver.was_stopped_early()          # True
solver.all_solutions()              # what finished, one entry per path
solver.solution_metadata()          # which paths succeeded, were stopped, or never began
solver.solve()                      # resumes: finished paths recalled, the rest tracked
```

The exception does not take the state with it, the solution accessors are derived lazily from the per-path slots, and `NeverStarted` already existed to mean what it says, so the four outcomes were all expressible before this; what was missing was any way to produce two of them.

## Tests

C++: a stop request before a solve leaves every path `NeverStarted` and the solver reporting it; withdrawing the request leaves the solver completely normal, so being stopped is not sticky; a bare tracker stops with no solver present.

Python, six tests, all delivering the stop from **inside** the solve via an observer that fires on the third completed path. A wall-clock timer was tried first and is the wrong tool: a 216-path solve finishes before the timer on a fast machine, the signal then hits Python's own handler after the solve's is gone, and `KeyboardInterrupt` is raised anyway, which passes and tests nothing. The observer fires at a point in the computation rather than at a time, so it is the same on every machine. The tests use `signal.raise_signal` rather than `os.kill` because it behaves identically on every platform the suite runs on. Covered: Ctrl-C raises and leaves the solver inspectable with the taxonomy adding up and Python's handler restored; a plain Ctrl-C after the solve still reaches Python; a programmatic stop returns without raising; a stopped solver solves to completion afterwards, which is the test that exposed the recall gap; a stale request made with nothing running is discarded; and the homotopy solver stops too.

Whole C++ suite, 997 passed in pytest, the tutorial doctests, both doc lints and the whitespace lint are green.

## Second commit: a path that did not succeed says where it got to

A failed path used to report a time of zero and no point, whatever the failure mode. Now every path's metadata carries the stamp. `final_time_used` is written after every stage, not only on endgame success. `last_point` holds the tracker's point when it gave up, filled only for a path that did not succeed, since a successful path's endpoint is its solution and there is no reason to duplicate it or let it wander into the solution slot. `num_successful_steps` and `num_failed_steps` count the predictor-corrector steps over the whole path, pre-endgame and endgame together, so the cost of a path is visible without timing it. `max_precision_used` is filled for fixed-precision solves too, where it read zero.

No observer is attached to get this. The tracker already holds its time, point and precision when `TrackPath` returns, and it already counts its steps in two integers it bumps on every step. Those reset at each `TrackPath` start, and an endgame issues hundreds per path, so the base tracker gets a second pair that only the caller resets: `ResetCumulativeStepCounts()`, `CumulativeSuccessfulSteps()`, `CumulativeFailedSteps()`. Two integer increments per step.

Found and fixed on the way: the endgame writes its approximation only when it converges, so a path whose endgame failed was handed the previous path's approximation as its solution. The slot is now emptied on failure, and the stamp is where its point lives.

The stamp travels through the MPI result, the records, and the Python binding, and the dataframe gains the three columns. Records written before this lack the fields and the decoder treats them as optional, the same way it already treats the singular values. Under the recall rule above the stamp matters for the wall-clock limit: it is the machine-independent account of how far an abandoned path got.

Tests: every path made to run out of steps partway is stamped with the code, a tally that adds up to the budget, a point of the right size, a time short of the boundary and the precision; a successful path carries the tally and no last point; a path that fails in the endgame is stamped and its solution slot is empty rather than stale; the records round-trip returns the stamp on recall; and two Python interface tests.

## Third commit: a wall-clock budget per path, and a recall policy

`max_wall_clock_duration` on the solver, in seconds with zero meaning none, gives every path a budget. A path that has not finished when it runs out is abandoned between steps with `SuccessCode.WallClockLimitReached`, stamped with where it got to, and recorded with the budget that stopped it. The check sits beside the stop check at the top of the tracking loop, so the overrun is at most one step. A bare tracker can be limited on its own. The primitive is a deadline rather than a duration, because an endgame issues hundreds of tracking calls for one path and a per-call budget would restart with each; the solver arms it before each path and clears it after.

The budget is deliberately not part of the run's identity. A budget says how long to wait for an answer, not what the answer is, so a 60 s and a 61 s run are the same ask, and encoding it would have bumped the config encoding version and invalidated every existing record for nothing. Wall-clock time is machine dependent; the recorded stamp is the machine-independent account.

What already-recorded work counts as done is now the user's decision rather than the library's, through `RecordsConfig.recall`, a `RecallPolicy`. `Nothing` tracks every path fresh, which is the old `recall = False`. `Completed`, the default, reuses completed paths and re-tracks abandoned ones, with one refinement: a path a wall-clock budget cut off is reused while the current budget asks no more patience than the one that abandoned it, and re-tracked once the budget goes up, which is exactly when paying again is the point. `Everything` reuses every recorded outcome as recorded. An interrupted path is always re-tracked, since an interrupt has no budget to compare. A bool is still accepted. The policy lives in its own small config rather than on the zero-dimensional config because the question is the same for every algorithm that records paths; it is added to the solver's config list, so `get_config` and `update` pick it up like any other, and the classic input parser has a stub for it since that format has no keyword.

The step and precision budgets need none of this: they are inside the ask already, so a different budget is a different run by identity and the old abandonment is never consulted.

Tests: a bare tracker with a deadline in the past stops at its first step boundary, having taken no steps, at its start point, and succeeds once cleared; through the solver a nanosecond budget abandons every path with the code, the stamp and the recorded budget, without the solve counting as stopped early and without a deadline left on the tracker, and an hour changes nothing; the records test walks the policy through a run under a nanosecond budget, the same budget again (reused), an hour under `Everything` (reused), an hour under `Completed` (re-tracked, and they finish), `Nothing`, and no budget at all against abandonments under one (re-tracked, since the ceiling moved to infinity); and six Python interface tests.

## Fourth and fifth commits: the whole-solve budget, ADR-0060, and a tutorial

`max_solve_wall_clock_duration` budgets the whole `solve()` call. The solver sets a deadline when the solve starts, never begins a path after it has passed, and arms the tracker with whichever of the path and solve deadlines comes first. The tracker cannot tell which deadline it hit, so when the solve deadline was the one that bound the solver rewrites the path's code from `WallClockLimitReached` to `ExternallyTerminated`, and the solve reads exactly as an interrupt: `was_stopped_early()`, `NeverStarted` and `ExternallyTerminated`, always re-tracked on recall. The per-path knob is renamed `max_path_wall_clock_duration` so the two read as a pair. The whole user-facing surface of this PR is two numbers of seconds and one enum, all off by default.

ADR-0060 records the decisions of the stage: one cooperative stop checked between steps and never a watcher thread or an exception across a worker; every path stamped, abandoned paths recorded and never-started ones not; budgets out of the ask; the reuse rule applied through the user's `RecallPolicy` in its own `RecordsConfig`; the whole-solve budget as an interrupt; the emptied solution slot of a failed endgame; new `SuccessCode` values appended last.

A doctested tutorial, "Stopping a solve, and giving paths a budget", under "Solver settings & multiprecision", covers Ctrl-C and `request_stop`, the stamp, both budgets, and how the recall policy reads an abandonment. Its budgets are a nanosecond and an hour so the outputs are the same on every machine, and the stop demonstration prints invariants rather than counts because the split between abandoned and never-started paths depends on the thread count.

## Not done, on purpose

Retrying only the failed or abandoned paths under different settings is an idea, not a plan; it is noted in the task list with the reasoning and nothing is built. The MPI solve is not covered by the interrupt or the budgets.

Closes #400.

That issue asked for a per-path budget producing an ordinary non-success code, and preferred a deterministic work budget, with wall clock acceptable as an opt-in safety valve on top, marked as making runs non-reproducible. What lands here is the safety valve in exactly that shape: opt-in, off by default, producing `WallClockLimitReached`, which the existing machinery treats as any other honest failure. The precision-weighted work budget is not built.

The reproducibility objection in the issue is answered concretely rather than waved past. The budget is excluded from the run's identity, so a truncated run and an untruncated one are the same ask and share their records. An abandoned path is recorded with the budget that abandoned it, so a replay can tell what it was. And recall re-tracks an abandonment as soon as the current budget exceeds the recorded one, so a replay on a faster machine, or with more patience, pays for the path rather than inheriting yesterday's truncation as an answer. What survives is the honest residue the issue was worried about: the same budget is the same amount of work only on the same machine, which the docs and ADR-0060 both state.


core/test/nag_algorithms/zero_dim.cpp
ofloveandhate, line 1: i think that there's overlap in the usage of NeverStarted

ofloveandhate: that is, NeverStarted can also mean that the path never got off the start time, because the start point was singular

ofloveandhate: I just checked, and I was wrong.  InitialRefinement clearly returns SingularStartPoint.  But I still want to do an audit on how to make things as clear as possible.

ofloveandhate: i believe i'm clear.

python_bindings/include/zero_dim_export.hpp
ofloveandhate, line 1: i think last_point will be a confusing name.  rename to latest_path_point

ofloveandhate:
i am using Claude to write this aspect of the library, and it's delightful.  Rest assured, I am reading the code, and doing my absolute best to produce a high quality product, mathematically correct, useful, pleasant to use.

PR #450, opened 2026-09-18
fix(parsing): report a declared-but-undefined name, and refuse a second pathvariable

### LINKED ISSUE #441: Classic parser does not enforce declaration counts: a declared-but-undefined function crashes with 'map::at'
There is a TODO on the classic parser's root rule (`core/include/bertini2/io/parsing/system_rules.hpp:135`):

    //TODO refine this so that counts are enforced at parse time?

The rule accepts any number of any declaration block, in any order, and never checks that what was declared was defined.  Probing it from Python found two real defects and several behaviours worth deciding on deliberately.

**A declared function that is never defined crashes with a meaningless message.**

    variable_group x;
    function f, g;
    f = x^2 - 1;

raises `IndexError: map::at`.  That is a raw C++ `std::map::at` lookup escaping through the bindings, which violates the rule that no C++ error reaches Python unexplained, and it tells the user nothing about the actual mistake, which is that `g` was declared and never given a definition.

**A second path variable is accepted silently.**

    variable_group x;
    pathvariable t;
    pathvariable s;
    function f;
    f = x*t;

parses without complaint.  A system has one path variable, so the second declaration should be an error rather than a quiet overwrite.

Behaviours that are probably correct but are worth confirming rather than leaving to accident: a definition for a name that was never declared is absorbed as a subfunction (Bertini 1 behaves this way, so likely right); a file with no functions, or with no variables, parses into an empty-ish system; a non-square system parses (correct, since randomization exists).  Declaring a constant or parameter and never giving it a value already produces a good parse error.

So the TODO is really two bugs and a question about how strict the parser should be.  The crash is the part that matters.

ofloveandhate:
The classic parser's root rule accepts any number of any declaration block, in any order, and checked nothing about what was declared. There was a TODO saying so. Probing it from Python turned up two real mistakes an input file could make without being told.

## A name declared and never defined

```
variable_group x;
function f, g;
f = x^2 - 1;
```

raised `IndexError: map::at`. That is a raw `std::map::at` escaping through the bindings. It names neither the forgotten name nor the problem, and a raw C++ error reaching Python is exactly what we do not want.

The cause is one line: `EmitDeclaredFunctions` walked the declared names and looked each one up with `definitions_.at(n)`, with nothing between a declaration and its definition. The lookup now goes through a small `DefinitionOf` that reports which name was declared and never defined, and what it was declared as:

```
[SystemParser] function `g` was declared but never defined.  Every declared name needs a defining line, as in `g = ...;`.
```

Constants and parameters already failed earlier, in the grammar, so their behaviour is unchanged; they are pinned by a test so that neither can reach `map::at` again either.

## A second path variable

```
variable_group x;
pathvariable t;
pathvariable s;
function f;
f = x*t;
```

parsed without complaint, quietly overwriting the first path variable. A system has exactly one, so the second declaration is refused.

The check is in the parser rather than in `System::AddPathVariable`, deliberately. The issue is about what an input file is allowed to declare. A caller building a system programmatically may still set the path variable and then change its mind, which `AddPathVariable` has always allowed, and there is no API for unsetting it, so making the setter throw would take that away to fix a parsing problem.

## What is left alone

The TODO is replaced by a note saying what is enforced and what is not. Declaration blocks stay order-free and count-free, because that is what Bertini 1 input looks like. A definition for a name that was never declared is still absorbed as a subfunction, which is also what Bertini 1 does.

## Tests

C++, in `classic_parsing_test.cpp`: a declared-but-undefined function names the culprit and the kind; a declared-but-undefined constant and parameter each still fail; two path variables are refused; one path variable still parses and the system has it. Python, interface only, in `parser_test.py`: the message that reaches Python names the culprit and no longer contains `map::at`, and two path variables raise.

Whole C++ suite, pytest, both doc lints and the whitespace lint are green.

Closes #441.


ofloveandhate (commented):
thanks!

PR #452, opened 2026-09-19
feat(python)!: one homotopy builder, straight_line_homotopy

### LINKED ISSUE #371: bertini.nag_algorithm.coefficient_parameter_homotopy should just be `straight_line_homotopy`
this is not specific to the coefficient parameter homotopy, it just glues together start and target in a regular old linear deformation homotopy

ofloveandhate:
`coefficient_parameter_homotopy` is not specific to coefficient-parameter homotopies. Its whole body was `make_homotopy(target, generic, path_variable, Integer(1))` — a straight-line homotopy with gamma pinned to 1. Next to it sat `blend_homotopy`, the same call with gamma passed through. Two names, one construction, differing in one argument.

So this does not rename one of them. It removes both and leaves the thing they were:

```python
straight_line_homotopy(target, start, *, path_variable='t', gamma=None)   # (1-t)*target + gamma*t*start
```

Read `blend_homotopy(a, b)` as `straight_line_homotopy(a, b)`, and `coefficient_parameter_homotopy(a, b)` as `straight_line_homotopy(a, b, gamma=1)`.

## Why those two names were the wrong ones

**"Blend" names the implementation.** A blend block is how a structured start system gets coupled in when node arithmetic cannot fuse it (ADR-0020). That is a fact about the evaluator, not about the homotopy, and it is equally true of the gamma=1 case, which also goes through it.

**"Coefficient-parameter" claims a specialization it does not have.** A coefficient-parameter homotopy is a straight-line homotopy whose start is a generic member of the family and whose gamma is 1, because the deformation is a path in parameter space and the gamma trick has no place in it. That is a choice of argument, and it now reads as one at the call site rather than being hidden in which function you called.

`gamma` accepts a plain int, so the parameter case is `gamma=1` with no `Integer` wrapper.

## Breaking, deliberately

Both old names are gone, with no deprecation shims, per the maintainer. Every call site in the repository moved: `parameter_sweep`, the records and zero-dim tests, the examples, five tutorials with their runnable twins, the namespace-migration note, and the error message that used to name both functions.

`parameter_sweep` now passes `gamma=1` explicitly. That is load-bearing and worth a reviewer's eye: it is a parameter-space sweep, so letting it fall through to the default would quietly put a gamma trick into a path that should not have one.

ADR-0020 keeps its decision — a structured-block start must be coupled by a blend, never by `System` node arithmetic — and gains a naming note recording this change. Its Context keeps the old name, since that is the history it is describing.

The C++ `MakeHomotopy` and its `make_homotopy` binding are untouched; `straight_line_homotopy` is the keyword-argument wrapper over it, as `blend_homotopy` was.

## Tests

No new behaviour to test, so no new tests: the existing ones moved to the new name, and the two that were named after the old functions now say what they actually check — that `gamma=1` does not drop a structured start, and that the `gamma=1` helper solves a parameter move.

991 passed in pytest, 178 doctests, the HTML docs build with no warnings, both doc lints and the whitespace lint are green.

Closes #371.


ofloveandhate:
Superseded by #453, which carries the identical commit from a branch named for what this is: a rename, not a new implementation.

Issue #457, opened 2026-09-19
path_truncation_threshold is settable but inert: the tracker is handed it in the same call that resets it


ofloveandhate:
## Summary

`TolerancesConfig::path_truncation_threshold` and `TolerancesConfig::newton_before_endgame` are settings a caller can read and write on a zero-dim solver, but there is no point in the lifecycle at which writing either one takes effect. The tracker keeps the default, whatever the config says.

## Why there is no working order

The tracker does not read these from the config. It is handed them once, positionally, by `DefaultTrackerSetup` (`core/include/bertini2/nag_algorithms/zero_dim_solve.hpp:1045`):

```cpp
tracker_.Setup(tracking::predict::DefaultPredictor(),
               this->template Get<Tolerances>().newton_before_endgame,
               this->template Get<Tolerances>().path_truncation_threshold,
               tracking::SteppingConfig(), tracking::NewtonConfig());
```

and `DefaultSetup` runs, in this order (`:919`):

```cpp
DefaultSettingsSetup();   // this->Set<Tolerances>(Tolerances());  -- resets to defaults
DefaultSystemSetup();
DefaultTrackerSetup();    // reads Tolerances, hands the values to the tracker
```

So setting the config **before** `DefaultSetup()` is discarded by the reset on the first line, and setting it **after** `DefaultSetup()` returns is never read, because the only read already happened. The two are consumed in the same call that destroys any value a caller supplied.

The classic-input path has the same shape at `:3483`.

## Evidence

```cpp
auto zd = algorithm::ZeroDimSolver<TrackerT, Cauchy, System>(sys);
zd.DefaultSetup();

auto tol = zd.Get<algorithm::TolerancesConfig>();
tol.path_truncation_threshold = 10;
zd.Set<algorithm::TolerancesConfig>(tol);

zd.GetTracker().InfiniteTruncationTolerance();   // 100000
```

Moving the three config lines above `DefaultSetup()` gives the same `100000`.

The working route is to bypass the config entirely:

```cpp
zd.GetTracker().SetInfiniteTruncationTolerance(10);
```

## Why it matters beyond the inconvenience

`path_truncation_threshold` decides when the tracker abandons a path as going to infinity. A caller who raises it because their solutions legitimately live far out, or lowers it to stop wasting time on divergent paths, gets no effect and no error — the solve behaves as though the setting were not there, and the only sign is the solutions that did or did not come back.

[...]

PR #458, opened 2026-09-19
feat(core)!: a system says which of its coordinates are auxiliary

### LINKED ISSUE #403: Make the finiteness classification customizable: is_finite judges all coordinates, including synthetic ones
## Summary

Finiteness is decided by the norm of the **whole** dehomogenized point, so an algorithm
that appends synthetic coordinates gets genuine solutions declared "at infinity" on the
strength of a coordinate that carries no geometric meaning.

```cpp
// core/include/bertini2/nag_algorithms/zero_dim_solve.hpp:2068
static_cast<NumErrorT>(user_pt.template lpNorm<Eigen::Infinity>()) <= post.endpoint_finite_threshold;
```

I would like to be able to tell the classifier **which coordinates finiteness applies
to**.

## Why the synthetic coordinate's magnitude is not a meaningful quantity

Several standard constructions solve a system whose variables are the natural ones plus
synthetic ones:

- **critical-point systems in null-vector form** — `f; v^T M; patch`;
- **doubled / preimage charts** — `(x, y)` with `y` on a second component;
- parameter and coefficient homotopies, and anything carrying a path variable into the
  endpoint classification.

Take the null-vector case.  `v` carries **its own patch** — necessarily, or the system
would admit the zero null vector — and that patch fixes a normalization nobody chose on
geometric grounds.  So `v`'s DIRECTION is the meaningful object and its MAGNITUDE is an
artifact of the patch: not intentionally divergent, simply **not governed**.  It varies
the way a random projection's scale varies, and is about as controllable.

That is what makes this a kind-error rather than a tuning problem:

- picking a better threshold cannot be right, because there is no correct threshold for
  an ungoverned quantity;
- constraining `v` cannot be right either — the patch is doing necessary work;
- excluding it from the question is the only thing that is right.

It also compounds with scale: the natural block's magnitude is set by the projection
draw, the synthetic block's by its patch, and an absolute threshold is applied to the
larger of two independently uncontrolled scales.

## Evidence

Measured while decomposing surfaces (a cellular-decomposition client of b2).  A critical
point whose natural coordinates were of order 1e3 — well inside the working region — was
classified non-finite because the null-vector block exceeded the threshold.  The client
carries a rescue for exactly this: it re-refines the natural part on the polynomial
null-vector system and recovers the point.  That rescue is bolted onto one call site,
for a verdict that should not have been rendered anywhere, and every other consumer of
this pattern will have to reinvent it.

## Possible shapes

Roughly in order of how much they'd help, all backwards compatible (default = today's
behavior):

1. **Coordinate selection in `PostProcessingConfig`** — indices, or a mask, saying which
   coordinates the finiteness norm is taken over.
2. **Per-variable-group by default** — judge finiteness on the natural variable
   group(s); systems built with a single group are unchanged.
3. **A caller-supplied predicate** — most flexible: "finite in x, don't care about v".

(1) or (2) suffices for the case above.  (2) is attractive because the grouping already
exists on the system, but note that some clients build these systems with a single
variable group deliberately (a precheck workaround), so a group-based rule needs a way
to say which group is natural.

## Scope

Deliberately limited to the finiteness **classification**.  The two path-truncation
thresholds have the same all-coordinates character, and the tracker's is independently
configured from the endgame's — filed separately so this stays one issue.

ofloveandhate:
Bertini judges a point by its largest coordinate, three times: the tracker truncates a path past `path_truncation_threshold`, the endgame abandons one past `Security::max_norm`, and the solver calls an endpoint past `endpoint_finite_threshold` infinite. Realness is a fourth question of the same shape.

All four measured the whole dehomogenized point. A critical-point system in null-vector form, `f; v^T M; patch`, must patch `v` or admit the zero null vector — and that patch fixes a normalization nobody chose, so `v`'s direction is the meaningful object and its magnitude is an artifact. Judged on it, genuine solutions were declared at infinity, and no threshold would have been right, because the quantity is not governed.

## What this adds

A system says which of its coordinates are auxiliary, by variable group or by index, and answers the questions itself:

```cpp
SetAuxiliaryVariableGroups({1});      // or SetAuxiliaryCoordinates({3, 4, 5})
IsFinite(point, threshold)
IsReal(point, threshold)
```

`InfinityNormOfDehomogenized` already described itself as "the single canonical how-big-is-this-point measurement", but only the endgames called it — the tracker and the classifier each hand-rolled the same expression beside it. All four sites go through the system now, so the declaration is taught in one place and needs no parameter anywhere: the system is what knows its coordinates.

Thresholds stay where they are. What was missing was never a setting; it was that nobody owned the judgement.

Realness had the identical defect one line from finiteness, so a point whose only complex coordinate was an ungoverned one read complex. Fixed in the same pass.

## Identity

The declaration is part of the system's content identity. The tracker truncates on the coordinates that are *not* auxiliary, so two systems differing only here do not compute the same thing and must not recall each other's records. This edits `b2sysenc/2` in place under ADR-0061, which is that rule's first use — 3.5 gets one encoding rather than one per change. The canonical decoder reads the two new lines, so the round trip still closes on digest equality.

## The bug underneath

`System`'s copy constructor is hand-written, member by member (`system.cpp:83`), and never mentioned the new members. Since `Clone` is a copy, the solver's target and its homotopy both arrived with an empty set: the feature worked perfectly on the object in the caller's hand and did nothing whatever inside a solve. Every System-level test passed while this was true.

It is a third hand-maintained mirror beside `serialize` and the canonical encoder. Fixed, commented as such, and pinned by a test. ADR-0062 records that a member added to `System` now has four places to go.

## Also found

`path_truncation_threshold` cannot be changed through the solver at all — `DefaultSetup` resets the configs and then hands the value to the tracker in the same call, so there is no moment at which a caller's value survives. Filed separately as #457; not touched here. The test that needed a lower threshold goes through `GetTracker().SetInfiniteTruncationTolerance` and says why.

## Shapes turned down

**A caller-supplied predicate.** A function cannot join the ask, so the same question would recall a differently-classified answer with nothing in the record able to explain it — #420's flaw, deliberately widened.

**Judging the first variable group by default.** In a multihomogeneous system every group is geometric. The grouping says where the patches are, not which coordinates carry meaning.

**A config read by all three layers.** The first design here. The System is the better owner: it already knows its coordinates, where a config would have needed plumbing to tracker, endgame and solver separately while meaning the same thing in all three.

## Tests

Twelve at the System level — each shape, the union, realness, surviving homogenization and patching, copies, the homotopy, identity, ordering-independence, the encoding round trip, both refusals. One on a bare tracker, where the same path is abandoned without the declaration and completes with it. One whole solve, from the system the caller wrote through the solver's clone, homogenize, patch and homotopy. Eight Python interface tests.

Whole C++ suite, 1031 pytest, 190 doctests, both doc lints, whitespace, both MPI legs, and the html docs build with warnings as errors: green.

A tutorial, "Auxiliary coordinates", with executed examples.

Closes #403.


ofloveandhate:
this will enable some nice workflows and code simplifications in other things i am working on.  nice.

PR #459, opened 2026-09-19
fix(endgames)!: one operating-zone criterion, shared by both endgames

### LINKED ISSUE #402: InEGOperatingZone is emitted unconditionally by the power series endgame (a genuine criterion in Cauchy)
`InEGOperatingZone` means two different things in the two endgames — unconditional in power series

The event is emitted on a genuine criterion in Cauchy and unconditionally in power
series, so a consumer cannot tell what it received.

**Cauchy** (`cauchy.hpp:993`) emits it once, *after* the c/k stabilization loop:

```cpp
while(!CheckForCOverKStabilization(c_over_k) && abs(ps_times.back()-target_time) > ...)
{
    auto advance_success = AdvanceTime<ComplexT>(target_time);
    ...
    c_over_k.push_back(ComputeCOverK<ComplexT>());
}
NotifyObservers(InEGOperatingZone<EmitterType>(*this));
```

That is a real operating-zone test: the cycle-number estimate has settled, so the
Puiseux asymptotics are believed to dominate. It fires once, after several advances,
and the advances before it are genuinely pre-zone.

**Power series** (`powerseries.hpp:651`) emits it at the end of *every* successful
`AdvanceTime`, with no test:

```cpp
SuccessCode tracking_success = this->EndgameTrackPath(next_sample, ...);
if (tracking_success != SuccessCode::Success) return tracking_success;
if constexpr (IsAdaptivePrec) if (...) return SuccessCode::HigherPrecisionNecessary;
NotifyObservers(InEGOperatingZone<EmitterType>(*this));   // every advance
```

So in power series it means "advanced", not "in the zone".

Power series has `ComputeCycleNumber()` and `upper_bound_on_cycle_number_` but no
stabilization test equivalent to `CheckForCOverKStabilization`, so there is currently
nothing for it to condition on.

**Visible symptom:** `GoryDetailLogger` (`observers.hpp:92`) logs "made it to the
endgame operating zone at time ..." for the event, so a power-series run logs that
line on every single time advance.

**Why it matters beyond logging.** The event is the natural marker for "the path is
now in the asymptotic regime", which is what a consumer needs in order to use the
endgame's samples quantitatively — e.g. fitting `log sigma` against `log |t|` to see
which singular values vanish at the root, where the model `sigma ~ |t|^(1/c)` only
holds inside the zone. Rungs from outside it do not lie on the power law, so a
consumer that trusts the event gets a silently wrong fit under power series and a
correct one under Cauchy.

Either the power-series emission should be conditioned on a real criterion (the same
c/k stabilization idea, which `CheckForCOverKStabilization` already implements), or it
should not be emitted there — but the two flavors should not disagree about what the
event asserts.

ofloveandhate:
`InEGOperatingZone` announces that a path has reached the asymptotic regime, where the Puiseux model the endgame is built on dominates. That is what makes the samples usable quantitatively: a consumer fitting a power law against `|t|`, to see which singular values vanish at the root, needs to know which rungs lie on it.

The Cauchy endgame emitted it once, after its c/k stabilization loop — a real test. The power series endgame emitted it at the end of every successful advance, with no test at all, so there it meant "advanced". A consumer could not tell which it had received, and got a silently wrong fit under one flavor and a correct one under the other.

## A lift, not a new criterion

`ComputeCOverK` reads only the geometrically spaced approach samples and a fixed random probe. The power series endgame's `samples_` is that same window under a different name — it is what `pseg_samples_` was named after — so the machinery lived in `cauchy.hpp` for historical rather than mathematical reasons.

`GetCOverKProbe`, `ComputeCOverK` and `CheckForCOverKStabilization` move to the base endgame. `ComputeCOverK` takes the samples as a parameter instead of reaching for one flavor's member, and the power series endgame announces the zone once, on the same test, keeping the same sliding window Cauchy keeps so the test reads the most recent estimates rather than the first ever taken.

The probe stays fixed for the endgame's life, which the move preserves. A fresh random probe per call made consecutive estimates disagree by probe noise alone, which could certify the zone spuriously.

## Settings

`minimum_for_c_over_k_stabilization` and `num_needed_for_stabilization` move from `CauchyConfig` to `EndgameConfig`, which is where a setting every endgame reads belongs. A script that set either on the Cauchy config needs a one-line change.

## Identity

Both encoders change, so every config digest moves. Under ADR-0061 this rides the existing `b2cfgenc/4`: 3.5 is unreleased, so the registry's top line is edited in place and no new token is minted. This is the second identity-affecting change to land under that one version, which is the rule doing its job — before it, this would have been `b2cfgenc/5` or `/6`.

The golden fixture is regenerated. The composition column is untouched, since no config joined or left the settings text.

## Tests

Each flavor announces the zone exactly once over a run that advances many times — the assertion that fails loudest if the criterion is dropped again. Then the criterion on its own: estimates that settle, estimates that wander, and a threshold the caller moved.

Cauchy's existing c/k tests now pass their samples explicitly. They already had the samples in hand, so it reads better than reaching into the endgame for them.

Whole C++ suite, 1023 pytest, 187 doctests, both doc lints, whitespace, both MPI legs, and the html docs build with warnings as errors: green.

Closes #402.


ofloveandhate:
this inconsistency had bothered me for a decade.  closed.

PR #460, opened 2026-09-20
the tracker's settings are configs, and final_tolerance has one owner

### LINKED ISSUE #457: path_truncation_threshold is settable but inert: the tracker is handed it in the same call that resets it

ofloveandhate:
Closes #457.

Four settings sat outside the config surface. Three were bare members of the tracker -- the predictor, the tracking tolerance and the path truncation threshold -- each with a hand-written setter and getter. The fourth, `final_tolerance`, was the opposite problem: a field of two structs at once, the solver's `TolerancesConfig` and the endgame's `EndgameConfig`, kept in step by a push at solve time.

Everything built on the config surface therefore had a hole in it. A settings bundle carried neither the tracking tolerance nor the predictor. The settings digest did not see them, so two solves differing only in predictor were the same ask. The generated configuration reference could not list them, because no amount of introspecting config structs finds a method. `path_truncation_threshold` had no route through a solver at all -- the only way to change it was to fetch the tracker and call a method, which is how it went unnoticed that a solver could not set it.

## What changed

The three tracker settings are `tracking::TrackerConfig`, in every tracker's `NeededConfigs`. `solver.update(predictor=..., tracking_tolerance=..., path_truncation_threshold=...)` works, a settings bundle carries them, the digest sees them, and the reference page lists them.

`final_tolerance` belongs to `EndgameConfig` alone. `TolerancesConfig` keeps the two Newton tolerances, which are the solver's to drive. No field name is shared by two configs any more, so the flat router has nothing to disambiguate. A solver-level `final_tolerance=` still works and reaches the endgame's config; the push and the `pushed_final_tolerance_` bookkeeping are gone.

The named setters and getters stay and read and write the config.

## Why the derived state moved

The bespoke setters existed for a reason: each also refreshed state derived from the setting. `SetTrackingTolerance` recomputed the digit count implied by the tolerance; `SetPredictor` rebuilt the predictor object and re-read its order. `Configured::Set<T>` is a dumb store, so a setting moved into a config and left with that design would take effect through one route and not another.

So the tracker rederives instead of caching. `SyncSettings()` runs at the top of `TrackPath` and rebuilds the predictor object only when the choice actually changed; the tolerance's digit count is computed on demand. A value arriving through `Set<TrackerConfig>`, through a restored bundle, or through the flat router is as effective as one arriving through a setter.

## Identity

Both moves change the settings text a solve folds in. 3.5.0 is unreleased, so `b2cfgenc/4`'s registry line is edited in place with its new keyspace hash rather than superseded, which is exactly the case ADR-0061 covers. The golden digest fixture moves with it.

## Classic format

`odepredictor` and `pathtruncationthreshold` parse into the new `TrackerConfig` rather than into `TolerancesConfig`; `finaltol` parses into `EndgameConfig`. The accepted input is unchanged.

## Not in scope

The partition is still uneven -- `NewtonConfig` and `TrackerConfig` both describe how the tracker takes a step -- and four tracker settings remain method-only (`infinite_truncation`, `precision_preservation`, `reinitialize_initial_step_size`, and ADR-0060's wall-clock deadline, which is deliberately out of the ask). Repartitioning is deferred; this is about where a setting lives at all.

## ADR

ADR-0063 records the two guardrails: a setting with no config is invisible to the digest, to bundles and to the reference, all three silently; and derived state cached in a setter goes stale, because three other routes replace the config without calling it.

## Verification

Gate battery all green locally: build, doclint, ctest 10/10, pytest 1027 passed 1 skipped, sphinx doctests, MPI both legs. Whitespace lint and the Python doclint pass.

New tests. The C++ ones are about the hazard the change had to avoid -- each sets a value only through the config, by a route no setter sees, and asserts the tracker behaves as if a setter had been called. The predictor case runs the same path three ways and requires the config-set run to do the same work as the setter-set run and different work from the one the tracker was set up with, so it fails if the predictor object goes stale. The tolerance case requires the derived digit count to follow. The truncation case requires a path running to infinity to be truncated at the config's threshold. At the solver: a predictor and a truncation threshold set on the tracker survive `DefaultSetup` and a solve, and `final_tolerance` has one owner that nothing reverts. `settings_test` covers the classic keywords landing in the config that owns them.

Python, interface only: both surfaces are one value, and a solver can reach the truncation threshold.


ofloveandhate:
a long-needed revision.

Issue #347, opened 2026-07-15
TypeError when using symbolic complex numbers.


ElderStack:
I was struggling to get the error to reproduce without just rewriting the whole program unfortunately. I included SliceMover at the bottom.

TypeError: No registered converter was able to produce a C++ rvalue of type Eigen::Matrix<boost::multiprecision::number<boost::multiprecision::backends::mpc_complex_backend<0u>, (boost::multiprecision::expression_template_option)1>, -1, 1, 0, -1, 1> from this Python object of type numpy.ndarray

```
import bertini as b2
import numpy as np
import import_ipynb
from SliceMover import SliceMover

def regeneration_base_points(sys, variables, eqn_index=0, rng=None):
    if rng is None:
        rng = np.random.default_rng()

    N = len(variables)
    degree = list(sys.degrees())[eqn_index]
    target_fn = sys.function(eqn_index)

    def rc():
        return complex(rng.uniform(-1, 1), rng.uniform(-1, 1))

    # (N-1) fixed generic slices, and degree generic linear factors
    # coefficient rows [a_1, ..., a_N, a_0] (variable coefficients, then constant term).
    slice_coeffs  = [[rc() for i in range(N + 1)] for i in range(N - 1)]
    factor_coeffs = [[rc() for i in range(N + 1)] for i in range(degree)]


    S = np.array(slice_coeffs)
    base_points_numeric = []
    for row in factor_coeffs:
        row = np.array(row)
        M   = np.vstack([row[:-1][None, :], S[:, :-1]])       # N x N
        rhs = -np.concatenate([[row[-1]], S[:, -1]])            # N
        base_points_numeric.append(np.linalg.solve(M, rhs))


    def bcoef(z):
        return b2.symbolics.Complex(str(z.real), str(z.imag)) #<-- symbolics complex here

    fixed_sys = b2.System()
    fixed_sys.add_variable_group(b2.VariableGroup(list(variables)))

    start_moving = b2.system.clone(fixed_sys)
    end_moving = b2.system.clone(fixed_sys)

    fixed_sys.add_linear_forms([[bcoef(c) for c in row] for row in slice_coeffs])

    start_moving.add_products_of_linears([[[bcoef(c) for c in row] for row in factor_coeffs]])

    end_moving.add_function(target_fn)

    def to_bertini_point(pt):
        return np.array([b2.symbolics.Complex(str(v.real), str(v.imag)) for v in pt]) #<--  symbolics complex here

[...]

PR #462, opened 2026-09-20
start-point precision, point-vector seams, and one homotopy builder

### LINKED ISSUE #348: Crash when using complex_mp
Crash when using complex_mp instead of symbolics complex. Same code as previous error, just switched the complex types.

```
import bertini as b2
import numpy as np
import import_ipynb
from SliceMover import SliceMover

def regeneration_base_points(sys, variables, eqn_index=0, rng=None):
    if rng is None:
        rng = np.random.default_rng()

    N = len(variables)
    degree = list(sys.degrees())[eqn_index]
    target_fn = sys.function(eqn_index)

    def rc():
        return complex(rng.uniform(-1, 1), rng.uniform(-1, 1))

    # (N-1) fixed generic slices, and degree generic linear factors
    # coefficient rows [a_1, ..., a_N, a_0] (variable coefficients, then constant term).
    slice_coeffs  = [[rc() for i in range(N + 1)] for i in range(N - 1)]
    factor_coeffs = [[rc() for i in range(N + 1)] for i in range(degree)]


    S = np.array(slice_coeffs)
    base_points_numeric = []
    for row in factor_coeffs:
        row = np.array(row)
        M   = np.vstack([row[:-1][None, :], S[:, :-1]])       # N x N
        rhs = -np.concatenate([[row[-1]], S[:, -1]])            # N
        base_points_numeric.append(np.linalg.solve(M, rhs))


    def bcoef(z):
        return b2.complex_mp(str(z.real), str(z.imag)) #<-- complex_mp here

    fixed_sys = b2.System()
    fixed_sys.add_variable_group(b2.VariableGroup(list(variables)))

    start_moving = b2.system.clone(fixed_sys)
    end_moving = b2.system.clone(fixed_sys)

    fixed_sys.add_linear_forms([[bcoef(c) for c in row] for row in slice_coeffs])

    start_moving.add_products_of_linears([[[bcoef(c) for c in row] for row in factor_coeffs]])

    end_moving.add_function(target_fn)

    def to_bertini_point(pt):
        return np.array([b2.complex_mp(str(v.real), str(v.imag)) for v in pt]) #<--complex_mp here

    base_points = [to_bertini_point(p) for p in base_points_numeric]

    return {
        'base_points': base_points,
        'base_points_numeric': base_points_numeric,
        'fixed_sys': fixed_sys,
        'start_moving': start_moving,
        'end_moving': end_moving,
        'degree': degree,
    }


#driver
#system setup
x, y = b2.Variable('x'), b2.Variable('y')
vars2 = b2.VariableGroup([x, y])
f1 = x - y + (x**2)
f2 = x - y + (y**2)

sys1 = b2.System()
sys1.add_variable_group(vars2)
sys1.add_functions([f1, f2])

result1 = regeneration_base_points(sys1, [x, y], eqn_index=0)
print(f"degree(f1) = {result1['degree']}  ->  {len(result1['base_points'])} base-level start points")
for p in result1['base_points_numeric']:
    print(" ", p)

sm1 = SliceMover(
    result1['fixed_sys'], result1['start_moving'], result1['end_moving'],
    result1['base_points'], new_gamma=b2.symbolics.Complex('0.6', '0.8'),
)
codim1_witness_points_sys1 = sm1.move_slice()

print("codimension-1 witness points on {f1 = 0} \u2229 {1 fixed slice}:")
for p in codim1_witness_points_sys1:
    print(" ", np.array(p))
```

### LINKED ISSUE #351: linear product total degree start points waste time computing in full precision
analogous to #346

### LINKED ISSUE #382: homotopysolver needs an option to autohomogenize/projectivize
because sometimes (mostly?) the user wants to not have to manually think about projectivization, and get the benefit of infinity being a finite point.  only rarely (all-real tracking for instance) does a user want to disable this functionality

### LINKED ISSUE #461: The first start point after a precision change is short of digits, silently
## What happens

The first multiprecision start point generated after a change of working precision is computed against patch state still at the *previous* precision, so it comes back at the new precision carrying only the old precision's worth of correct digits. The second and later points at that precision are correct.

Found while testing #351, in `MHomogeneous`, whose target system in the test is homogenized and patched. `TotalDegreeLinearProduct` over an unhomogenized target shows nothing, which is what points at the homogenize-and-patch lift rather than at the linear solve.

## Measured

Largest residual of the start system at its own start points, over two consecutive passes at the same working precision, precision raised between rows:

| working precision | first pass | second pass |
|---|---|---|
| 30 | 3.9e-31 | 4.1e-31 |
| 45 | 1.7e-31 | 2.8e-46 |
| 60 | 1.9e-46 | 1.7e-61 |
| 90 | 1.6e-61 | 2.9e-91 |
| 120 | 7.1e-92 | 4.3e-121 |

The first pass at each precision is accurate to the *previous* row's precision. The second pass is accurate to the row's own. At 120 digits the first pass is short by 29 digits.

The same table for `TotalDegreeLinearProduct` on an unpatched target is clean at every row (1.2e-30, 1.4e-45, 1.6e-60, 1.5e-90, 2.4e-120), first pass and second alike.

## Why it looks like this

`GenerateStartPointT` solves the linear system in affine coordinates and then calls `HomogenizePoint`, which inserts the homogenizing coordinates and rescales onto the patch. Evaluating the system afterwards brings the patch to the working precision, but nothing does so before the lift. So the ordering within one pass is: generate point 0 against a patch at the old precision, evaluate (patch now syncs), generate point 1 against a synced patch, and so on -- one degraded point per precision change, the first one.

Per ADR-0057 each holder of multiprecision values self-aligns to the precision of the point it is handed. The lift appears to be a path where that does not happen, or happens after the value it affects has already been used.

## Why it matters

Silent digit loss. The point is returned *at* the requested precision, so nothing downstream can tell it is short: `Precision(start_point)` reports the working precision, and only a residual check reveals the missing digits. Adaptive precision moves the working precision during a solve, so this is reachable outside a test.

## Not the same as #351

#351 was the linear solve running at the coefficient masters' storage precision and rounding afterwards -- wasted work, correct answer. This is the reverse: a cheap answer with fewer correct digits than advertised. The #351 fix neither causes nor cures it; the table above was taken with that fix in place.

### LINKED ISSUE #463: Homogenize() half-works on a system containing a blend block, and says nothing
## What happens

`System::Homogenize()` on a system that contains a blend block leaves the system inconsistent and unusable, silently. It homogenizes the ordinary blocks and adds a homogenizing variable to the variable group, but a blend block's `Homogenize` is a deliberate no-op, so the blend still describes the old, un-homogenized variable set. The system then reports one variable count and evaluates against another.

## Measured

Building the moving homotopy from `nag_algorithm.moving_homotopy` (a fixed block plus a blend of two moving systems) and homogenizing it:

```
before homogenize: num_variables = 2  num_functions = 2  is_homogeneous = False
  eval at a 2-vector: [-0.47-0.22j  0.4 -0.3j ]        <- fine

after homogenize : num_variables = 3  num_functions = 2  is_homogeneous = False
  eval at a 2-vector raised RuntimeError: number of variables doesn't match
  eval at a 3-vector raised RuntimeError: number of input variables (3) doesn't
                                          match number of system variables (2)
```

Neither length evaluates. `num_variables()` says 3, the blend still says 2, and `is_homogeneous()` stays False even though `homogenize()` was called and reported no error. A straight-line homotopy, which has no blend block, homogenizes correctly in the same probe and evaluates at 3.

## Why the no-op is right, and the silence is not

`blend_block.hpp` says it plainly: a blend is the coupling homotopy, built from operands that are already prepared, and its operands are `shared_ptr<const System>` and cannot be mutated. That is the design (ADR-0020, ADR-0026), and `ZeroDimSolver` follows it -- it homogenizes and patches the target *first*, then builds the homotopy from it.

The defect is that calling `Homogenize()` on such a system is accepted and half-performed rather than refused. A caller who does it gets a system that passes construction, reports a plausible variable count, and then fails at evaluation with a message about variable counts that does not say what they actually did wrong.

## Suggested

Refuse: `Homogenize()` on a system containing a block that cannot homogenize should throw, saying that a blend's operands are fixed at construction and that the operands must be homogenized before the blend is built. Leaving the system untouched is the other defensible answer; what it must not do is half the job.

## Context

Found while looking at #382 (an auto-projectivize option for `HomotopySolver`). Projectivizing after the homotopy is built works for a straight-line homotopy and cannot work for a blend one, which is what points at doing it in the homotopy builders instead.

### LINKED ISSUE #464: num_functions() counts the patch row, function(i) does not return it
## What happens

For a patched system, `num_functions()` includes the patch row but `function(i)` refuses the index that row sits at:

```
IndexError: System::Function index out of range
```

So the obvious loop -- ask for each function over `range(system.num_functions())` -- raises on the last one.

## Why it surfaced now

It was reachable before, since any patched system has it, but rarely met: the systems a caller asks `function(i)` about tend to be the ones they authored, which are not patched. #382 makes the homotopy builders projectivize by default, so patched systems now turn up where affine ones used to.

## Suggested

Pick one and make the pair agree: either `function(i)` returns the patch row at the index `num_functions()` counts, or the count excludes it and something else exposes it. The first keeps `num_functions()` meaning the length of the vector an evaluation returns, which is what a caller iterating is usually after.

ofloveandhate:
Closes #351. Closes #348. Closes #382. Closes #461. Closes #463. Closes #464.

Six issues, in the order they were found: each one after the first was turned up by the work on the one before it, which is why they are together.

## #351 — linear-product start points computed a thousand digits to keep thirty

`TotalDegreeLinearProduct` and `MHomogeneous` each produce a start point by solving an n-by-n linear system whose coefficients come from masters stored at `MaxPrecisionAllowed` — a thousand digits, so the masters are precision-faithful. Boost's variable-precision backend computes at its operands' precision, so the whole LU ran at a thousand digits and the rounding at the end of the routine discarded every digit past the working precision.

The masters now enter the solve already at the working precision. They are themselves untouched, so asking for the same start point at a higher precision still refines toward the same point rather than a different one.

Measured over three passes of every start point:

| case | before | after |
|---|---|---|
| 3 variables, degree 3 (27 points) | 0.015s | 0.004s |
| 4 variables, degree 2 (16 points) | 0.014s | 0.002s |
| 2 variables, degree 3 (9 points) | 0.003s | 0.000s |

The tell is that the old timings did not depend on the working precision at all — 30 digits and 60 digits cost the same, because the cost was set by the masters' storage precision rather than by the caller's choice. This is the follow-up #346 left behind when it fixed the same shape in the binomial start system; #351 names the linear-product one, and `MHomogeneous` had the identical lambda one file over.

The regression test asserts what a cheaper solve could break rather than a time: the points are still roots to the accuracy the working precision affords, more precision still buys accuracy, and the high-precision point agrees with the low-precision one to the low precision.

## #461 — the first start point after a precision change was short of digits, silently

That test found a second defect. Rescaling a point onto a patch read the patch's *working* coefficients without first bringing them to the precision of the point. Every other evaluable path self-aligns that way (ADR-0057); this one demanded the caller align it beforehand, in an assertion a release build compiles away — so the one check that could have caught it was absent exactly where it mattered.

A rescale following an evaluation at a lower precision therefore returned a point *reporting* the requested precision while carrying only the older one's correct digits. Generating start points does exactly that ordering. Measured before the fix, largest residual over two consecutive passes at the same precision:

| working precision | first pass | second pass |
|---|---|---|
| 45 | 1.7e-31 | 2.8e-46 |
| 90 | 1.6e-61 | 2.9e-91 |
| 120 | 7.1e-92 | 4.3e-121 |

At 120 digits the first point is short by 29, and nothing downstream could see it: `Precision()` reported 120, and only a residual check revealed otherwise. Adaptive precision moves the working precision during a solve, so this was reachable in ordinary use.

## #348 — multiprecision values in an object-dtype numpy array

The crash reported there is gone: the reported program shape solves cleanly over 200 random coefficient draws, and the fix appears to be 12f280ed, written against #347 and landed nine hours after #348 was filed. Both reports are one root cause through two types — a numpy array the native converter cannot take.

What was still wrong is the seams one layer over. `np.array` over `complex_mp` values infers the registered `complex_mp` dtype, which the converter accepts; the same values in an array built with `dtype=object`, grown by assignment, or sliced out of a larger object array sit behind a dtype no overload matches. `HomotopySolver` rebuilds its points and was safe. `System.eval` and `Tracker.track_path` did not, and answered with a dump of C++ Eigen signatures — verified for `track_path` by removing the conversion and watching the error return.

Both take an object array of numbers now, and a plain list, exactly as start points already did. The conversion lives in one private module rather than a copy per seam. An object array of anything else is still refused rather than reinterpreted, so this widens what works and changes nothing that did.

## #382 — the homotopy builder projectivizes, and there is only one of it

A user building a homotopy wants one that will solve the problem, and that means working over projective coordinates, where infinity is an ordinary place a path can reach instead of somewhere it runs off to. `ZeroDimSolver` has always done this with the systems it builds for itself.

The builder is the only place it can happen: once a blend block exists its operand systems are fixed and cannot be homogenized, so projectivizing a finished homotopy is not available. `ZeroDimSolver` is not special in being a solver — it is special in being a *builder*.

So `moving_homotopy` is gone, and `straight_line_homotopy(target, start, fixed=None)` does both jobs, because they were always one construction: with no `fixed` the whole system deforms, with it only the rest does. The held rows are what the old name was reaching for and naming badly — nothing moves; some rows simply carry no path-variable dependence and are evaluated once rather than blended, contributing exactly zero to dH/dt. The moving rows may be complete systems, a `Slice`, or a plain list of functions when `fixed` supplies the variable structure.

It returns a `StraightLineHomotopy`: `.homotopy`, `.target`, `.start`, `.fixed`, `.gamma`, `.path_variable`. An inert record — it remembers and does nothing else, holds a System rather than being one, and has no identity of its own, so records still key on the homotopy System as before. A general `Homotopy` type is deliberately not what this is: the invariants a specific construction can state are exactly the ones an arbitrary family cannot.

Neither end system is the caller's to assemble any more, which is a correctness matter rather than a convenience — the rows must come held-first and nothing would have caught it otherwise. `.start` is the system whose solutions are the start points, `.target` is what the answers satisfy, and `.gamma` is the coefficient actually used, random default included, so a path is reproducible from the record. `HomotopySolver` and `bertini.solve(homotopy=...)` take the record directly and the target comes with it.

Projectivizing works on clones. A System's content is its identity and records key a solve on it, so converting a caller's system in place would silently change what their own system *is* and stop an earlier solve of it being recalled — measured, while trying exactly that. Solutions already come back in user coordinates (ADR-0013), verified on a projective homotopy against an affine target, so the data flows in the caller's coordinates in both directions. A family that cannot be written projectively — a products-of-linears block over more than one affine group — is built affinely rather than half-converted.

`projectivize=False` builds the homotopy exactly as written, which is what you want when the affine coordinates are the point, as in all-real tracking.

## #463 — homogenizing a system that contains a blend half-worked, and said nothing

Trying to projectivize a finished homotopy is what turned this up. A blend's `Homogenize` is deliberately a no-op, since its operands are `shared_ptr<const System>` and fixed at construction (ADR-0020, ADR-0026). But the surrounding system homogenized its *other* blocks and added a homogenizing variable to the variable group, leaving the parts disagreeing about how many variables there were: `num_variables()` reported 3, the blend still said 2, `is_homogeneous()` stayed false, and evaluation then refused both 2 and 3 coordinates — with a message about variable counts that did not name the mistake. It refuses now, and says where homogenizing belongs.

Combining systems that describe points differently is refused for the same reason: a homotopy between a projective system and an affine one deforms between points that do not correspond, and nothing downstream notices. The variable-structure check `System+=System` already made is now `CheckVariableStructuresMatch`, which `MakeHomotopy` and `MakeMovingHomotopy` make too. Homogenizing-variable count is checked first, so a projective/affine mix is diagnosed as such rather than as a variable count that happens to differ.

## #464 — a patched system's patch rows are functions, and reachable as such

Projectivizing by default made patched systems ordinary, which exposed this. `num_functions()` counts the rows an evaluation returns, patch rows included; `function(i)` returned only the functions as authored, so the two disagreed by exactly the number of patches and the obvious loop over `range(num_functions())` raised on the last index.

Once a system is patched the patch *is* one of its functions — a row of the vector `eval` returns and a row of the Jacobian — so it is reachable through `function(i)` now, built from the same walk of the variable ordering that `Patch::EvalInPlace` and the symbolic Jacobian already make. `functions()` still returns the equations as authored and nothing else, so copying them into another system copies equations rather than somebody else's choice of patch; `get_functions()` is the whole evaluated set. The Python docstring for `num_functions` had also said it excluded patches while being bound to the count that includes them.

## Also here

`bertini.max_precision()` is exposed alongside `default_precision()`. Drawing the default gamma at the adaptive ceiling needs it: a multiprecision node caps at its creation precision, so a gamma minted at the working precision would pin the path there.

## Breaking

`moving_homotopy` is removed rather than deprecated, and `straight_line_homotopy` returns a record rather than a bare System. Both break loudly. The cellular and NID work in the port calls the old builder and will need repointing.

## Verification

Gate battery green on the committed tree: build, doclint, ctest 10/10, pytest 1043 passed 1 skipped, 191 sphinx doctests, MPI both legs. Whitespace lint, the Python doclint, and an HTML docs build with no warnings.

The moving-slice tutorial, its companion script, and `everyday.rst` are on the new spelling; the tutorial gained a section on why the homotopy is projective and what that looks like.


ofloveandhate:
it was a lot of fun to work through the challenge of unifying a few methods for making homotopies, which had grown a lot in June and July.  I now have StraightLineHomotopy, and a better canonical way of making them.  I believe this will let the interface stabilize!!!!

PR #468, opened 2026-09-21
printing objects in Python

### LINKED ISSUE #99: add __str__ methods for trackers, endgames, and other things that we want to inspect.

ofloveandhate:
Closes #99.

106 of the 188 public classes printed `<bertini._pybertini.tracking.AMPTracker object at 0x7f3c...>`: the type you already knew, and an address you cannot use. That was every tracker, every endgame, all 38 observer event types, the observers and collectors, the linear algebra decompositions and the decomposition scaffolding.

The count overstates the work. Most of those are one template exposed at three precisions, so about fifteen functions cover all 106.

## What it reads like now

```
<AMPTracker: homotopy in 1 variable, RKF45, tolerance 1e-05, adaptive precision at 30 digits, at t=0.31 after 47 steps>
<PrecisionIncreased: 30 -> 40 digits, at t=0.7>
<SuccessfulStep: at t=0.65 in 40 digits>
<PathDataCollector: 12 steps>
<SolutionPathCollector: 4 paths, 65 steps>
<AMPPowerSeriesEndgame: from t=0.1, to t=0, not yet run>
<PartialPivLU: 3x3>
<MidpathCheckReport: passed, 0 crossings>
<GoryDetailLogger>
```

## Three decisions

**The text lives in the python layer**, in `bertini/_repr.py`, attached at import: a Boost.Python class accepts the assignment, and `str()` follows `repr()` for any type that does not define its own. A repr is presentation, and presentation for python readers is its own question — a C++ `operator<<` serves a different audience and can be written when one is wanted, without this having pre-empted its shape. Keeping it in one file also means the whole convention can be read at once.

**Machinery is described in angle brackets; values keep the call form.** None of these objects could be rebuilt from any text, and most hold a C++ object that no python expression can name, so `AMPTracker(...)` would imply a round trip that does not exist. Configs, metadata and `StraightLineHomotopy` already read `SteppingConfig(initial_step_size=..., ...)`, which is close to something you could type back, and they keep it.

**A tracker's description is live.** It reports where the path is now rather than only how it was configured, because the moment you most want to print a tracker is inside an observer callback mid-path.

Nothing here may raise. Every accessor goes through a helper that swallows failure and omits that clause, since they all reach into C++ state that may not exist yet. A repr that throws turns an ordinary `print` in a debugging session into a traceback about printing.

ADR-0065 records all of it, with the guardrails.

## Three defects found on the way

**A tracker's step counters were read uninitialized.** `ResetCountersBase` runs at the top of `TrackPath`, so until a path had been tracked, `num_total_steps_taken()` read two members that had no initializers — on a fresh tracker it returned whatever had been in that memory, which is how it was noticed (1093431419). The two cumulative counters, added later, already had initializers; the four per-call ones now do too.

**A variable group printed the addresses of its variables.** `repr` gave `[0xae8010b19520, 0xae8010b100f0]` while `str` gave `[x,y]`, which is how it went unnoticed: the bound container streams its elements in C++, and an element is a pointer to a node. It reads `VariableGroup([x, y])` now, as does a list of them.

**A `SolveResult` named things it did not have.** With recording off there is no run and no directory, and the repr said so as `SolveResult(4 solutions, run , 0 recalled, records at None)`, which reads like something went wrong when nothing did.

## The regression test

`python/test/classes/repr_test.py` walks every module under `bertini`, collects every public class, and fails if any of them inherits the default `__repr__`. A binding added without a description fails there rather than reaching a user. The remaining tests check that the descriptions are true rather than merely present: that a tracker's reported position matches where the path ended, that a collector's count matches what it collected, and that a fresh object's repr neither raises nor contains an address.

A type that genuinely has nothing to report still names itself — `<GoryDetailLogger>`. Several of these expose no state at all, which is a gap in what the bindings expose rather than in this file; when such a type gains accessors, its description should gain the clause.


ofloveandhate (commented):
Dope. Glad the DLL stuff gets hidden into `_`.  And glad that now printing will be much more useful.  Not the way I would have implemented it myself if left to my own devices -- I would have it scattered all over the place, honestly.  I like that it ended up mostly int one file.  We'll see what happens in the long term.  Glad to have this.

ofloveandhate:
waiting for Windows CI to be green before merging, since we changed the DLL stuff on a non-Windows dev machine..

Issue #469, opened 2026-09-21
progress reporting needs to stay in C++ at very large path counts


ofloveandhate:
The progress display added in #359 is a python observer: the solver emits `PathStarted` and `PathComplete` for every path, and the observer counts them. That is the right shape at the path counts b2 runs today, and the wrong one at the counts it is heading for.

The redraw interval already scales -- it defaults to about two hundred updates across the whole solve however many paths there are, which is what b1's `printpathprogress` setting was for. What it does not scale is the **callback**. Two python callbacks happen per path whether or not anything is drawn, because the events are emitted either way and the observer is attached in python.

At a few thousand paths this is immaterial. At ten million it is twenty million trips across the binding, and at three hundred million it is six hundred million -- on the order of ten minutes of pure overhead, spent deciding not to draw.

The fix is a progress observer written in C++, attached on the C++ side, that never enters python at all: it keeps its own counters, and draws (or hands a summary to whatever is reporting) on its own throttle. Python would ask for it by name rather than supply the callback.

Worth settling when it is built:

- Whether it draws directly, or emits a coarse "progress" event that python can render -- the second keeps the display in one place but reintroduces a callback, just a thousand times rarer.
- What it does under MPI, where today only the manager rank reports.
- Whether the same mechanism serves the log for an unattended run, which is a different need from a bar and is not well served by anything today.

Not urgent. Filed so that when a solve of that size is actually being run, the reason it feels slow is already written down.

PR #470, opened 2026-09-21
a solve says how far it has got

### LINKED ISSUE #359: need progress bars / reporting during solving
because it's painful to not know how far a solve is.

ofloveandhate:
Closes #359.

A solve told you nothing while it ran. Now it says how far it has got:

```
paths:  61%|██████     | 39/64 [00:41<00:26]  37 ok, 2 diverged
```

It rides the path events the solver already emits, so the solver still knows nothing about display.

## On by default

The complaint is that a solve is silent while it runs, and a switch you have to discover does not answer that for the people who feel it. What makes the default tolerable is that the display shows itself only when it would help:

- **Nothing for the first five seconds.** The many small solves inside a loop, a tutorial or a test finish before a bar would appear.
- **Nothing at all off a terminal.** tqdm's own `disable=None` turns it off when stderr is not a TTY, so scripts, pipelines, captured output and CI see none of it. It writes to stderr rather than stdout for the same reason.
- **One rank under MPI.** Only the manager reports; the workers would otherwise each draw their own.

`show_progress=False` on any solve is the certain silence.

## Recalled paths are reported separately

A solve that finds its answers in the records runs no paths and emits no path events, so a bar would sit at zero and then vanish — the one behaviour that looks like a hang. Measured, with the seed pinned so the ask genuinely repeats:

```
first  : 9 solutions, 0 recalled
second : 9 solutions, 9 recalled
events on the recalled solve: {'AlgorithmStarted': 1, 'AlgorithmComplete': 1}
```

So the bar sizes itself to the paths actually being tracked, and the rest are announced on their own line: `recalled 9 of 9 paths from the records`.

## Throttling, and what it does not buy

Redrawing on every path is the problem b1's `printpathprogress` setting existed to solve. Here the interval defaults to whatever gives about two hundred updates across the whole solve, at any path count, and can be set exactly.

What that does **not** save is the callback: two python callbacks happen per path whether or not anything is drawn, because the solver emits the events regardless. That is immaterial at thousands of paths and would be about ten minutes of pure overhead at three hundred million, where the answer is an observer that stays in C++. Filed as #469 rather than built now.

## The C++ half

**`AnyAlgorithm` is `Observable`.** Being watchable is a property of being an algorithm, rather than something each one remembers to add: a caller holding nothing but an `AnyAlgorithm` can attach an observer without knowing which algorithm it has or how it is templated. Virtually inherited, since the concrete algorithms reach that base through their own type-erased bases (`AnyZeroDim`, `AnyNID`) and must not end up with two observer lists — the direct `Observable` bases on `HomotopySolver` and `NumericalIrreducibleDecomposition` were exactly that second copy and are gone. A test attaches through the base, detaches through the base, and checks the second solve is silent.

**`PathComplete` carries how the path ended.** The code is already written to that path's own metadata slot at both emission points, so carrying it costs nothing, and it lets a watcher tally outcomes as they arrive instead of reading metadata that a worker thread is still filling in. A test checks that what the event said about a path is what the solver recorded for it. Note that `Success` there is not the same as a finite solution: finite, at infinity or singular is settled after every path is done.

**`num_paths()`**, the count a solve is working through, known once setup has run — which is before the first path event, so a watcher can size itself from `AlgorithmStarted`.

## Threading

Path events are emitted from inside `ExecuteOnePath`, which runs on the worker pool. Measured with `num_threads=4`: the callbacks arrive on five distinct python thread ids, nothing crashes, and every path is counted exactly once (a test pins that). C++ serializes notification under the observable's own mutex.

## tqdm

A dependency now, rather than forty lines of bar drawing maintained here. The three places CI installs the wheel with `--no-index` gain it explicitly, since that install resolves nothing on its own.

## Docs

The progress section went into the "stopping a solve, and giving paths a budget" tutorial, because a bar that stops advancing is how you *see* a frozen path, and `max_path_wall_clock_duration` is what gets the solve past it. The two belong on the same page.


ofloveandhate:
The observer system I implemented  in the core pays off over and over and over.   noninvasive status  bars?!?   yes, please.

Issue #215, opened 2026-05-16
provide wheels for macos-intel


ofloveandhate:
There's a hard crash in unit testing in python.  I was unable to fix it -- relying on CI for it is too laggy and distant.  Once I get an intel machine again (crossing fingers) i can test myself in real time and fix.  Until, intel wheels not built.

ofloveandhate:
I think the crash might be from conversion from np.int64 to multiprecision complex???

ofloveandhate:
Intel macOS wheels are built and tested again, as of #471 and #472.

The crash is gone, and the guess in this thread was the right one. The full pytest suite now passes on `macos-15-intel` on every python version — `1074 passed, 2 skipped` on py3.11, and green on 3.10, 3.12, 3.13 and 3.14 — where it used to SIGABRT. Nothing was done here to address the crash directly: the eigenpy boundary was rebuilt over the summer (uninitialized mp slots healed on read and zeroed before assignment, guarded ufunc loops and casts, numpy slots written only through `slot_write`; ADR-0006 and the work after it), and "conversion from np.int64 to multiprecision complex" reads like exactly that family. By the time intel was asked again, there was nothing left to crash.

What #471 had to fix was three things that would have failed an intel runner for reasons of their own, each of which would have produced a red run that told us nothing:

The prebuilt Boost+eigenpy bundle was named `deps-macos14-arm64-...` unconditionally, so an intel runner's download *succeeded* and unpacked arm64 dylibs, which then failed to link — an architecture mismatch wearing the costume of a build error. The asset name carries the architecture now.

Homebrew lives at `/opt/homebrew` on apple silicon and `/usr/local` on intel, and only the first was named.

The wheel-test job ran only when the matrix contained the literal `macos-14`, so intel would have built wheels that nothing tested — the quietest possible way to answer this question wrongly.

#471 also gave intel the same prebuilt and cached state the other unix runners get: the C++ job's 30-minute backstop was cancelling a cold first build, and a cancelled job skips its post steps, so the ccache was never saved and every subsequent run was cold too; the wheel ccache was keyed on `runner.os`, which is `macOS` for both architectures, so the two evicted each other inside one budget; and the deps builder only ever built arm64.

[...]

PR #471, opened 2026-09-21
re-enable intel in CI

ofloveandhate:
Toward #215. **This PR is an experiment: its own CI run is the result.** Merge it if the intel jobs come back green, close it if they do not — either way we learn something we have not been able to learn since the runner was dropped.

Intel macOS left the matrix because pytest died there with SIGABRT, and the crash was never reproduced locally, there being no intel mac to hand. The comment in the workflow blamed "an MPFR/Boost interaction"; the note on #215 suspected "conversion from np.int64 to multiprecision complex".

That second guess reads like the family of faults the eigenpy boundary has since been rebuilt around: uninitialized mp slots healed on read and zeroed before assignment, guarded ufunc loops and casts, numpy slots written only through `slot_write` (ADR-0006 and the work that followed). The same misattribution happened once already on Linux, where a crash was blamed on the container's MPFR version and turned out to be version-independent. If intel was the same story, there is nothing left to crash.

## Three things that would have failed intel for unrelated reasons

None of these are the crash; all three would have produced a red run that told us nothing.

**The dependency bundle is per-architecture, and the name was not.** `CIBW_BEFORE_BUILD_MACOS` fetched `deps-macos14-arm64-...` unconditionally. On an intel runner that download *succeeds* and unpacks arm64 dylibs, which then fail to link — an architecture mismatch wearing the costume of a build error. The name carries `uname -m` now, so intel 404s and takes the from-source path, which is the right thing until an intel bundle exists.

**Homebrew is not at the same place on both architectures.** `/opt/homebrew` on apple silicon, `/usr/local` on intel, and both the eigenpy cmake invocation and `CIBW_ENVIRONMENT_MACOS` named only the first. Both are named now; a prefix that is not present is skipped.

**The wheel-test job checked for the literal `macos-14`.** Intel would have built wheels that nothing tested — the quietest possible way to answer this question wrongly, and the one that would have looked like success.

## What to expect

The intel leg builds Boost and eigenpy from source, so those jobs will be slow — this is the fallback path, not a regression.

If it SIGABRTs again, the crash is at least reproducible on demand now: take `macos-15-intel` back out of the matrix line and keep `-f os=macos-intel`, which runs intel alone at a single python version, so the question can be asked again cheaply. A stack trace from that run would be worth more than the four guesses on the issue.


ofloveandhate:
It was successful, and this PR also gives the ccache speedup to intel.

